PR #69, opened 2014-03-14
Integrating PhantomJS

sgarbesi:
I started integrating PhantomJS into the crawler then stopped myself because I didn't want to hack the code up or do it in a way that you may have not preferred. There are a number of ways you can go about doing this, the preferable way in my opinion would most likely be as a separate function; but since I am not as familiar with the code as you, the author, it's probably best that you go about that part of the integration.

If you have any questions let me know. Running a simple exec() with the command below and waiting on the callback for stdin or stderr should suffice.

**An example of how to run this with PhantomJS:**
phantomjs lib/phantom.js --useragent="Test Agent" --timeout=10000 --url=http://angularjs.org

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -94,6 +94,10 @@ var Crawler = function(host,initialPath,initialPort,interval) {
 	// Internal cachestore
 	crawler.cache			= null;
 
+	// Use PhantomJS?
+	crawler.phantom			= false; // Whether or not PhantomJS should be used for requests.
+	crawler.phantomBin		= 'phantomjs'; // Binary location of PhantomJS.
+
 	// Use an HTTP Proxy?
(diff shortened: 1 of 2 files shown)


defkrie:
@sgarbesi 
how to use phantomjs with your commit ?
crawler.phantom			= false; // Whether or not PhantomJS should be used for requests.
i set crawler.phantom to true, but i have not change...

sgarbesi:
@defkrie It's not directly tied into the library, I was leaving that part up to the author. See my original comment with the example.

cgiffard:
After much deliberation, I decided that integration was not really in simplecrawler's domain. Instead, it would be better to consume it in another package in order to add phantom support. Thanks so much for the PR though — I'm really sorry for the hassle.

PR #74, opened 2014-05-07
Add config option to follow initial redirects to a different domain.

tobli:
If crawler.allowInitialDomainChange is set to true, and the initial url redirects to a different domain (e.g. a .com redirects to a canonical .net domain), crawler.host is updated to the new domain, to allow crawling to continue on that domain.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -249,6 +249,9 @@ Here's a complete list of what you can stuff with at this stage:
 	`Node/SimpleCrawler <version> (http://www.github.com/cgiffard/node-simplecrawler)`.
 *	`crawler.queue` -
 	The queue in use by the crawler (Must implement the `FetchQueue` interface)
+*   `crawler.allowInitialDomainChange` -
+    If the response for the initial url is a redirect to another domain, update `crawler.host` to
+    continue the crawling on that domain. Defaults to false.
 *	`crawler.filterByDomain` -
 	Specifies whether the crawler will restrict queued requests to a given
 	domain/domains.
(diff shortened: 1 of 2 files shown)


cgiffard:
Hi — sorry about being unimaginably slow about reviewing this! If you're still interested in pulling it in, could you please add some tests? Thanks!

tobli:
No worries, I'll be back with a test. Thanks!

tobli:
Hi @cgiffard
After thinking about how to test requests to different domains I've come to the conclusion that the best way is to use http mocking using https://github.com/pgte/nock. I'll introduce that as a test dependency for my test, unless there's a smarter way I haven't thought of.

cgiffard:
I used nock for about a year in the tests for a large a production app before, and I think I'd recommend you stay clear of it — it tends to make tests very fragile. In the end for our app we decided on DI. Unfortunately simplecrawler was not written to support DI, so that's not possible. A simpler solution might be to export the http.request function for mocking... I might think about how to refactor it to make testing easier.

bahout:
Hi
We need this feature too. Is this feature implemented?

shaunakv1:
Hello, doesn't look like pull request has been merged. We need this feature very much too. Any plans on merging this to the master? Thanks!

fredrikekelund:
To me, it looks like this PR kind of fell between the cracks. @tobli, would you be interested in opening a new one? The changes look good to me. As for the test, I think we can get away with a lighter solution than nock. I would recommend adding a route that redirects to 0.0.0.0 (default server operates on 127.0.0.1). That should give us a different host, but all the same routes. In any case, the test could probably be completed in a "fetchstart" call, since we don't really care about what happens to the request, we only want to know that the crawler respects the redirect.

tobli:
Thanks @fredrikekelund 
I'll open a new PR and will look into your suggested testing approach. It would be awesome to get this in.
ping @soulgalore

PR #111, opened 2015-03-06
Adds regression test to: Referer undefined validation - in requestOptions #107

pose:
Related to #107


cgiffard:
Thanks, this looks good.

pose:
Thanks for merging @cgiffard! Do you think you can release a new version to npm? Thanks!

PR #119, opened 2015-04-12
Update Travis config.

XhmikosR:
Use node.js 0.12, 0.12 and io.js.

It seems node-simplecraweler has issues with node.js > 0.10  https://github.com/ChrisWren/grunt-link-checker/issues/12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: node_js
 node_js:
-  - "0.11"
   - "0.10"
+  - "0.12"
+  - "iojs"
+matrix:
+  fast_finish: true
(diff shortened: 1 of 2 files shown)


cgiffard:
Thanks! Looks good. :)

XhmikosR:
Thanks! I was hoping to find a solution for https://github.com/ChrisWren/grunt-link-checker/issues/12 but it doesn't seem to be an issue with node-simplecrawler itself :/

Maybe you have have an idea @cgiffard?  https://github.com/ChrisWren/grunt-link-checker/blob/master/tasks/link-checker.js

cgiffard:
@XhmikosR Try the latest version? I think the header bug should be fixed now.

PR #120, opened 2015-04-15
Option to add query parameters to every request

sabrehagen:
I needed the ability to send a query parameter along with any url my crawler requested from my server. In my case this crawler is used to generate a cache of prerendered pages for my angular application. By passing the refresh=true query parameter, I clear the cache for the requested url and save a new version of the page to the cache.

There was no way to supply query parameters with each request, so I have added it in this PR. The format is an object with key value pairs for each query parameter and its value.

If you would like any variable names changed  to fit convention let me know and I can change them. There is also no test for this. There is no test for the querystring strip either, so I didn't put one in for this. Let me know if you would like me to write one.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -271,6 +271,8 @@ Here's a complete list of what you can stuff with at this stage:
 	Or go even further and strip WWW subdomain from requests altogether!
 *	`crawler.stripQuerystring` -
 	Specify to strip querystring parameters from URLs. Defaults to false.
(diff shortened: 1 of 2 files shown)


cgiffard:
Hey — looks mostly good! Couple of things:

1) Could you add a test/s? :)
2) Instead of string mashing, it might be best to use URI.js to add the query string parameter — it's already available and ready for you to use.

Let me know if you want to talk it over! :)

PR #132, opened 2015-05-05
Ignore Depth Regex

BenChirlin:
I don't know if you want to fold this into the main repo but I don't see why not as it only adds and option and otherwise does not affect the repo. I've added an additional option to both the code base and docs that allows a user to specify a regex that, if matched by a `queueItem` will be crawled no matter but what the current depth.

I needed this for the project I was using the repo for since I had posts paginated across multiple pages where I wanted the crawler to explore and scrape all the pages. So, for instance, I have this option set to ignore depth on all requests that end with `?pg=` some number to ensure I get the entire article even if the first page is queued when depth is already at max.

Hope that makes sense and thanks for the amazing library!!!

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -295,6 +295,10 @@ Here's a complete list of what you can stuff with at this stage:
 
 *	`crawler.discoverRegex` -
 	Array of regex objects that simplecrawler uses to discover resources.
+*	`crawler.discoverNoDepthRegex` —
+	Array of regex objects that simplecrawler will explude from	`maxDepth`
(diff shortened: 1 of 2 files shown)


cgiffard:
Thanks for the PR! I think I'd like to rationalise discovery — there's a hell of a lot of really specific requirements and the logic around whether something should be queued or not is getting very complex. It might be better to rationalise the API to enable developers to easily add their own logic.

BenChirlin:
That makes sense. Just put a callback filter function in as an option. Thanks!

PR #137, opened 2015-05-15
fixed bug(host and subdomain is not being replaced)

raintean:


cgiffard:
Not sure how that one slipped through.

It's an obvious one, so I'm going to merge it, but if you file another PR, please provide tests! :)

PR #142, opened 2015-05-27
filterByDomain check should only be checked when adding to queue

doug-nara:
crawler.filterByDomain should only restrict what we add to the queue, not what is considered "valid". will suggest in another commit an additional configuration that controls whether we restrict discovery to valid hosts


cgiffard:
Hi — thanks for submitting a PR, and sincere apologies for neglecting it for so long!

If you're still interested in this feature, I'm happy to help merge it in if you could add some tests.

doug-nara:
I would love to but I’m not sure I’ll get back to this project in any sort of timely fashion.

PR #149, opened 2015-06-17
Stall out on 404's fix

dylanPowers:
I was having issues where the crawler would stall out when I hit 404's or anything that wasn't a 200. By "stall out" I mean that the process.nextTick callback would not be called for a lengthy period of time with nothing else appearing to run. Adding socket.end() fixed it for some reason. Anyone know why? I couldn't find any documentation in node that explained what was going on.

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -1066,7 +1066,6 @@ Crawler.prototype.handleResponse = function(queueItem,response,timeCommenced) {
 
 			response.on("data",receiveData);
 			response.on("end",receiveData);
-
 		} else {
 			response.socket.end();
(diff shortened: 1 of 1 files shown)


cgiffard:
This is an interesting one. I can't see the harm in merging it for now, even though you haven't provided tests (I thought I'd let it slide since I'm not sure how to actually write a test for this, haha.)

I wonder whether this relates to (depending on the node version) the number of HTTP agents — if the socket is keepAlive, perhaps it hangs waiting for them to drop out (and the TTL on those sockets might be 60 seconds or five minutes.)

Issue #140, opened 2015-05-22
The web-programming curse of the Irish: M'name is O'Neal m' boy.


BrandonCouts:
So like a O'Neal.jpeg terminates the url detection early. Upon inspection this is because the regex doesn't try to match the opening quote with an identical closing quote. I added another discover rule to help solve this problem, though this is at the cost of more processing. I'm not that clever with regex, but I reckon this can be turned into a single rule again.

```javascript
	crawler.discoverRegex = [
		/(\shref\s?=\s?|\ssrc\s?=\s?|url\()([^\"\'\s>\)]+)/ig,
		/(\shref\s?=\s?|\ssrc\s?=\s?|url\()["]([^"]+)/ig,
		/(\shref\s?=\s?|\ssrc\s?=\s?|url\()[']([^']+)/ig,
		/http(s)?\:\/\/[^?\s><\'\"]+/ig,
		/url\([^\)]+/ig,
		/^javascript\:[a-z0-9\$\_\.]+\(['"][^'"\s]+/ig
	];
```

BrandonCouts:
Also is there a reason why: 
/http(s)?\:\/\/[^?\s><\'\"]+/ig

Doesn't capture url's with queries in them?

BrandonCouts:
```Javascript
crawler.discoverRegex = [
	/(\shref\s?=\s?|\ssrc\s?=\s?|url\()([^\"\'\s>\)]+)/ig,
	/(\shref\s?=\s?|\ssrc\s?=\s?|url\()((["][^"]+)|(['][^']+))/ig,
	/http(s)?\:\/\/[^\s><\'\"]+/ig,
	/url\([^\)]+/ig,
	/^javascript\:[a-z0-9\$\_\.]+\(['"][^'"\s]+/ig
];
```
I was wondering if you might comment about maybe getting this into the source. It seems to me this would be pretty straight forward change? The discovery regex needs the "?" I believe so the crawler can stripe it later as well? What are your thoughts on this?

fredrikekelund:
@BrandonCouts can we close this?

BrandonCouts:
Yup

PR #154, opened 2015-08-06
Correctly discover URLs that legitimately end with a quote or parenthesis

aquavitae:
I discovered this problem while scraping a truly horrible site that was full or URLs of the form `"scripts/run.dll?12341:Project+(Year)"`.  This PR makes a two changes to address these types of URLs:

1. Change the `discoverRegex` list so that trailing quotes or parentheses are kept
2. Change the `cleanURL` function to correctly identify and extract trailing quotes or parentheses.

In the process, I noticed that `discoverRegex` might match the same URL multiple times.  I don't think this is necessarily an issue since duplicate queued items should be passed over, and parsing from the text is a bit of guesswork anyway.  However, I added a comment just to note it in case the issue ever comes up.

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -148,10 +148,14 @@ var Crawler = function(host,initialPath,initialPort,interval) {
 
 	// Regular expressions for finding URL items in HTML and text
 	crawler.discoverRegex = [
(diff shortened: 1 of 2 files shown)


BrandonCouts:
I've noticed this behavior as well. This solves my O'neal issue.
+1
#140

cgiffard:
Thanks! Looks great. Good tests make for an easy merge. :)

PR #156, opened 2015-08-23
Added alternative versions of ampersand html encodings

gillesdemey:
I stumbled across this in one of my projects, took me a while to figure out the server used different codes to encode HTML entities.

Added some tests to prevent regressions in the future.


cgiffard:
Great! Thanks for that. :)

PR #158, opened 2015-09-03
added some aditional checks for url path + superAgent pool

buonaparte:
Possible some tests must be added for the use case

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
+.idea
 /test/queue.json
 /npm-debug.log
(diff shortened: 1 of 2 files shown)


cgiffard:
Thanks for filing a PR. I definitely like reliability fixes — though the userAgent stuff is a more specific feature that I don't think necessarily has a place in SimpleCrawler core.

I do want to make an API that makes adding things like that easier though. If you could explain what you need, perhaps we could add some hooks for that use case.

I'm closing this PR for now.

If you'd like to open another one related to the path stuff, please do the following:

* Write tests for the feature
* Provide a description in the PR of the purpose of the feature/fix and how it works.

Thanks again. :)

PR #160, opened 2015-09-10
Expose cleanExpandResources method and trim discovered resources

fredrikekelund:
In some scenarios, developers might want to provice custom logic for extracting links, but still make use of simplecrawler's internal logic for cleaning and fixing up those links (eg. if you have a DOM parser library and only want to extract links from the href attributes of links). With this change you can call `crawler.cleanExpandResources` to do that!

Also now trimming all discovered resources, since URL's can't actually contain spaces and browsers etc. will ignore any spaces in href attributes.

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -426,6 +426,20 @@ Crawler.prototype.processURL = function(URL, context) {
     };
 };
 
+/*
+    Private: Perform string replace operations on a URL string. Eg. removes
+    HTML attribute fluff around actual URL, replaces leading "//" with
+    absolute protocol etc.
+
+    queueItem - Queue item corresponding to where the resource was found
+    URL       - String to be cleaned up
+
+    Examples
+
+        cleanURL({protocol: "http"}, "url('//example.com/about') ")
+
+    Returns a string.
+ */
 function cleanURL (queueItem, URL) {
     return URL
         .replace(/^(?:\s*href|\s*src)\s*=+\s*/i, "")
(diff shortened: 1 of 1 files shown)


fredrikekelund:
I'll look into the failing tests!

fredrikekelund:
:tada:

lib/crawler.js
```diff
@@ -412,6 +412,72 @@ Crawler.prototype.processURL = function(URL,context) {
...
+				URL = URI(URL)
+					.absoluteTo(queueItem.url || "")
+					.normalize()
+					.toString();
+			} catch (e) {
+				// But if URI.js couldn't parse it - nobody can!
+				return list;
+			}
+
+			// If we hit an empty item, don't add return it
```
cgiffard: s/add//

:)

lib/crawler.js
```diff
@@ -440,81 +505,18 @@ Crawler.prototype.discoverResources = function(resourceData,queueItem) {
 	if (!queueItem.protocol)
 		queueItem.protocol = "http";
 
-	if (!crawler.parseHTMLComments) {
+	if (!crawler.parseHTMLComments)
```
cgiffard: Couple of small things — I used to use this style, but @XhmikosR submitted a PR that removed it, and I more or less agree that we should standardise on braces now (especially as the project is maturing a little.) Could you add the braces back in?

(I wonder if we could stick jscs in there to make this check automatic.)

fredrikekelund: Done! I agree, adding some JSCS checking would be cool - good to have an enforced style

cgiffard: Thanks! Sorry for being a stick-in-the-mud. :)

cgiffard:
It might be beneficial to document this method, too. I'll look at putting something together tomorrow morning. :)

fredrikekelund:
No problem with the styling, I like to keep things consistent! Good idea to document it. Would you like me to fold the commits into a single one btw? Makes for a neater commit history

cgiffard:
I don't really mind for now. I think it'd be a bit silly to get hung up on the exact contributor process when I haven't finalised anything yet. Probably best to write some guidelines first. :)

fredrikekelund:
Alright, cool!

XhmikosR:
@fredrikekelund: please fetch and rebase (squash if needed).

fredrikekelund:
Sure thing, been planning to, will fix it tomorrow hopefully!

lib/crawler.js
```diff
@@ -412,6 +412,72 @@ Crawler.prototype.processURL = function(URL,context) {
 	};
 };
 
+var cleanURL = function (queueItem, URL) {
```
cgiffard: Perhaps this should have documentation too — just tagged with `Private:`.

fredrikekelund:
@cgiffard I've documented the cleanURL function and rebased everything. Think this is ready to be merged?

cgiffard:
Hit the button. :)

PR #166, opened 2015-09-16
Switch to using ESLint.

XhmikosR:
@cgiffard: the rules I find interesting but are off for the time being are those with `0` in .eslintrc. Feel free to push to this branch and we can rebase before merging.

Don't forget `?w=1` in the URL to see the diff without comparing whitespace.


example/phantom-example/index.js
```diff
@@ -42,7 +40,7 @@ crawler.emit = function(name, queueItem) {
         return string;
     }
 
-    if (!~boringEvents.indexOf(name)) {
+    if (boringEvents.indexOf(name) === -1) {
```
cgiffard: Thanks for fixing this! :)

Next time I'll try to do it myself so I don't have to bother you, haha.

cgiffard:
Looks great, thanks!

PR #167, opened 2015-09-16
Decrement _openRequests when mimetype isn't supported

### LINKED ISSUE #144: Crawler stuck indefinitely
I'm having an issue where the crawler gets stuck for an indefinite amount of time (until, in my code, it is manually aborted after a timeout of 5 minutes). Since every individual request has a configured timeout of 1 minute, it shouldn't be possible for the crawler to reach the 5 minute mark without any activity.

I suspect this is related to #136 and occurs after the crawler has encountered a number of timeouts (although I haven't seen any pattern where this problem always arises after X number of timeouts).

I tried temporarily downgrading to see if the suggestion that this problem was introduced somewhere around 40a71c1c5abffc2729316f4a90a0cb70b056612f might hold some truth, and it does seem like 0.3.11 doesn't have the same problem.

fredrikekelund:
Fixes #144. Big problem, small solution. The indefinite hangup of the crawler was due to the fact that I had flipped downloadUnsupported to false (and this problem was triggered in that case).

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -1124,7 +1124,9 @@ Crawler.prototype.handleResponse = function(queueItem, response, timeCommenced) 
             response.on("data", receiveData);
             response.on("end", receiveData);
         } else {
+            queueItem.fetched = true;
             crawler._openRequests--;
+
             response.socket.end();
         }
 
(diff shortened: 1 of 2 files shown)


fredrikekelund:
As a sidenote, it seems this issue was introduced in b00e91b5e56372900877d99374cf252944878805

cgiffard:
Oh, wow. *headdesk*

Sorry about causing you so much hassle — and thanks for finding the issue! I see you've got a good test there for replicating it. I'll try get it out to npm ASAP.

fredrikekelund:
Haha that's ok! Happy to find that the root of the problem was something easy to fix. Yeah, it'd be great if you could ping me once you've published!

PR #168, opened 2015-09-16
fixed parseScriptTags

### LINKED ISSUE #165: parseScriptTags not working properly
If the option `parseScriptTags` is set to false simplecrawler only crawls the first page and stops. I recognized this behavior with the Content Management System [Contao](https://contao.org/de/). It can be tested with the [demo](http://demo.contao.org/en/). Probably it is not a problem with Contao only, I couldn't figure out the problem yet. Maybe some different HTML Markup.

lgraubner:
fixed problem in regex to match zero or more instead of one or more
characters

this commit fixes #165

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -548,7 +548,7 @@ Crawler.prototype.discoverResources = function(resourceData, queueItem) {
     }
 
     if (!crawler.parseScriptTags) {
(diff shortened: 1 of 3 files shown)


cgiffard:
Hey! Thanks so much for finding the issue and adding a PR!

Would you mind adding a test for this feature? :)

lgraubner:
There you go. I hope the test is sufficient.

cgiffard:
That's great! Thanks. :)

PR #178, opened 2015-10-16
Prevent timeout events from stop crawler

gorbatenkov:
Hello, after commit from this request https://github.com/cgiffard/node-simplecrawler/pull/170 I found another problem with node 4.1.2 in clientRequest.setTimeout(crawler.timeout, function() {...});

In my test case when some redirects happen one by one socket already die, but clientRequest is still alive and when timeout event fired queueItem.fetched == true and we can't proceed to clientRequest.abort(); and crawl stoppes forever because _openRequests == maxConcurrency.
I've modified code to check clientRequest._crawlerHandled instead of queueItem.fetched and everything works fine on several projects

Here is example of problem project logs:
I've run project http://www.swannetoscar.com with maxConcurrency = 5
After 5 redirects _openrequest become 5 too and stoppes forevever
Fetch redirect: http://www.swannetoscar.com/factory/shirt/index/collar_product_id/918/ to http://www.swannetoscar.com/factory/shirt/#collar
Fetch redirect: http://www.swannetoscar.com/factory/shirt/index/collar_product_id/917/ to http://www.swannetoscar.com/factory/shirt/#collar
Fetch redirect: http://www.swannetoscar.com/factory/shirt/index/collar_product_id/919/ to http://www.swannetoscar.com/factory/shirt/#collar
Fetch redirect: http://www.swannetoscar.com/factory/shirt/index/collar_product_id/915/ to http://www.swannetoscar.com/factory/shirt/#collar
Fetch redirect: http://www.swannetoscar.com/factory/shirt/index/collar_product_id/916/ to http://www.swannetoscar.com/factory/shirt/#collar

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -918,8 +918,8 @@ Crawler.prototype.fetchQueueItem = function(queueItem) {
 
         clientRequest.end();
 
-        clientRequest.setTimeout(crawler.timeout, function() {
-            if (queueItem.fetched) {
+        clientRequest.setTimeout(crawler.timeout, function () {
+            if (clientRequest._crawlerHandled) {
                 return;
             }
 
@@ -927,9 +927,12 @@ Crawler.prototype.fetchQueueItem = function(queueItem) {
                 crawler._openRequests--;
             }
 
-            queueItem.fetched = true;
-            queueItem.status = "timeout";
-            crawler.emit("fetchtimeout", queueItem, crawler.timeout);
+            if (!queueItem.fetched) {
+                queueItem.fetched = true;
+                queueItem.status = "timeout";
+                crawler.emit("fetchtimeout", queueItem, crawler.timeout);
+            }
+
             clientRequest._crawlerHandled = true;
             clientRequest.abort();
         });


fredrikekelund:
This looks good to me. I just want to try and see if I can reproduce the issue, @gorbatenkov have you set any special configuration for your crawler or can I try it more or less with the default settings?

gorbatenkov:
I think that the only important configuration in my case is
projectCrawler.timeout = 60000;
Other settings was close to default one

gorbatenkov:
Forget to notice, that problem reproduce only with node 4.1.2, but not with previous version 0.10.40

fredrikekelund:
@gorbatenkov okay, so I tried to reproduce the problem with your configuration on that domain, and I can't seem to get the crawler to hang. I ran it using node 4.1.2 on Windows and simplecrawler 0.5.4 with the following [config](http://pastebin.com/0TtrrArB). Would you mind trying the same script to see if that works for you or not?

gorbatenkov:
Hello, okay I will try at the end of the day

gorbatenkov:
I can't reproduce problem with your code and after one more session after debug I found problem in other script of our project here:
var http = require('http');
http.globalAgent.maxSockets = 5;

With this settings you can see crawler hang - http://pastebin.com/fs93Wx7c

Interesting, that when I remove http.globalAgent.maxSockets = 5; from project crawler finished, but as in your example pages http://www.swannetoscar.com/factory/shirt/#collar, http://www.swannetoscar.com/factory/shirt/#wrist returns 404 error

cgiffard:
Those last 404s you're seeing appear to be a bug — we shouldn't be sending fragments up in URIs — the sanitiser should chop them off.

gorbatenkov:
And what about http.globalAgent.maxSockets situation? This is a problem or not?

cgiffard:
This looks like a bug — the crawler shouldn't hang like that. I can't reproduce the issue yet though... I'm going to try to create a small test for it.

gorbatenkov:
Hello, so what do you think after all? My commit is useless beacuse came from http.globalAgent.maxSockets = 5 situation?

fredrikekelund:
Yeah, I think we can probably skip it, so I'll go ahead and close it.

Doesn't look like it screws anything up either, and that kind of ambiguity might mean there's room for refactoring in the page fetching logic, but it feels odd to merge it if there isn't a problem that it fixes

PR #183, opened 2015-11-03
Added decodeResponses option

fredrikekelund:
With this enabled, the crawler looks for a character encoding in
the response headers and falls back to the first 512 bytes of the
response body, where it looks for a <meta> tag that specifies a
charset. This information is then passed to iconv-lite that does
the actual decoding.

By default this option is disabled.

I also added a test.


test/lib/routes.js
```diff
@@ -96,5 +98,9 @@ module.exports = {
 
     "/script": function(write) {
         write(200, "<script src='/not/existent/file.js'></script><script>var foo = 'bar';</script><a href='/stage2'>stage2</a><script>var bar = 'foo';</script>");
+    },
+
+    "/encoded": function(write) {
+        write(200, iconv.encode("test ströng with ä few non english chåracters", "iso-8859-1"), "text/html; charset=ISO-8859-1");
```
cgiffard: I'm wondering whether we should load in an external fixture file(s) with a few different encodings just to make sure that the text encoding of the file isn't resolved automatically by somebody's editor, etc.

fredrikekelund: Probably a good idea, this test case is pretty minimal right now, I'll see if I can expand on it a bit

fredrikekelund:
As a reminder to myself: we should add tests that detect the charset from an inline meta tag as well.

tomtom87:
+1

fredrikekelund:
@cgiffard what do you think - is this ready to be merged and can we push a new version then? Perhaps appropriate to bump to 0.6.0 because of this being a new feature?

cgiffard:
Unless I'm mistaken, the package.json is missing a reference to the iconv-lite dependency — but once that's resolved lets bump the version and merge. Let me know when you wanna push to npm! :)

fredrikekelund:
I think you might've just missed it actually, I can see it in the diff. Check abb122f3a2398b84abb896d23f7646d7f57654e4

cgiffard:
Oh, I did! My apologies. :)

fredrikekelund:
@cgiffard cool! You want to bump the version and publish to npm?

fredrikekelund:
@cgiffard could you do an `npm publish`? :) Would be great to have this feature available in simplecrawler

lib/crawler.js
```diff
@@ -958,6 +963,31 @@ Crawler.prototype.fetchQueueItem = function(queueItem) {
...
+    Returns the decoded buffer.
+
+*/
+Crawler.prototype.decodeBuffer = function (buffer, contentTypeHeader) {
+    contentTypeHeader = contentTypeHeader || "";
+
+    var embeddedEncoding = /<meta.*charset=["']{0,1}([^"'>]*)["']{0,1}\s*\/{0,1}>/i.exec(buffer.toString(undefined, 0, 512)) || [],
+        encoding = contentTypeHeader.split("charset=")[1] || embeddedEncoding[1] || contentTypeHeader;
+
+    encoding = iconv.encodingExists(encoding) ? encoding : "utf-8";
```
luozhihua, line 985: It seems that if `content-type =“ image / * ”` or any other response that is not a text file will be decoded into `utf-8`, in fact decode pictures, Word, Excel, PDF, ZIP and other files into` utf- 8` will cause the file to be damaged and cannot be opened.

PR #191, opened 2015-11-13
Customizable agent for http / https.

keichan34:
Also adds a quick SOCKS proxy example (requires the `socksv5` package)

I wasn't sure what the best way to test this was, so I'm open to suggestions. If creating a dummy agent that does nothing is non-trivial (node noob here), that might be an option.

### DIFF
--- a/example/socks-example.js
+++ b/example/socks-example.js
@@ -1,0 +1,33 @@
+var Crawler = require("../");
+var socks   = require("socksv5");
+
+var socksAuthNone = socks.auth.None;
+
+var socksSettings = {
+    proxyHost: "localhost",
+    proxyPort: 9050,
+    auths: [ socksAuthNone() ]
+};
+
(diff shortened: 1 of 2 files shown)


fredrikekelund:
I think it would be cleaner to skip the added arguments to the constructor and instead just let the developer replace `myCrawler.httpAgent` and `httpsAgent` respectively after having instantiated a Crawler.

Other than that the diff looks pretty good to me. As far as testing goes, I think it might be a good idea to provide a working example, but let developers add http agents "at their own risk". Thoughts, @cgiffard?

keichan34:
That's fine, I think 99% of the use cases won't need to be using an alternate agent anyways. I'll update the PR in a bit!

keichan34:
There we go. Also updated the example in `example/socks-example.js`

cgiffard:
That looks good! I'll update the documentation and version after merging.

PR #193, opened 2015-11-17
fix: determine port correctly for HTTPs resources

joscha:
Currently when using an HTTPs URL without explicitly specifying the port, the default port is incorrectly determined to be `80` when it should be `443`.

### DIFF
--- a/lib/quickcrawl.js
+++ b/lib/quickcrawl.js
@@ -63,7 +63,7 @@ module.exports = function crawl(url, successCallback, failCallback) {
         throw new Error("Can't crawl with unspecified path.");
     }
 
-    var tmpCrawler = new Crawler(url.hostname(), url.path(), url.port() || 80);
+    var tmpCrawler = new Crawler(url.hostname(), url.path(), url.port() || (url.protocol() === "https" ? 443 : 80));
 
     // Attach callbacks if they were provided
     if (successCallback) {


XhmikosR:
Can you squash those 2 patches into one?

joscha:
@XhmikosR done

joscha:
after looking through the code because of another port/protocol related error, I think we would also need to fix this in crawler.js and make the `initialProtocol` a constructor parameter in order to be able to start crawling an HTTPs endpoint. Are you okay with that?

fredrikekelund:
I think a cleaner solution would be to use urijs to parse a single string that goes into the constructor instead of chopping up what could easily be a single argument into more pieces.

Making that change straight off would kind off break compatibility though, so either that should effectively be version 1.0.0, or the constructor should only try to parse the first argument as a url if just one argument is passed. Since the first argument is the hostname, the results would still most likely be correct even if we always parsed the first argument as a url, but it feels convoluted to maintain two ways of doing such a trivial thing. 

@cgiffard, you have an opinion? Since it concerns bumping the major version number

cgiffard:
Lets raise all the big architectural/API issues separately to interim fixes against the 0.X versions, and mark them as API overhaul for 1.0. In the meantime I think we can merge this in! :)

fredrikekelund:
Good call! Thanks @joscha :)

cgiffard:
I'll probably raise an epic for the 1.0. Lots to go over. :)

PR #196, opened 2015-11-25
Tiny things.

bfwg:
I noticed some little things that the code can do better?
Maybe I'm wrong.
For the emit part I tried the code below to test: 
```js
const util = require('util');
const EventEmitter = require('events');

function MyEventEmitter() {
  EventEmitter.call(this);
}

util.inherits(MyEventEmitter, EventEmitter);

var a = new MyEventEmitter();



a.on('test', (obj) => {
  obj.test = 'something';
});

var obj = {test: ''};
a.emit('test', obj);


console.log(obj);// => {test: 'something'}

```

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -252,20 +252,20 @@ Crawler.prototype.start = function() {
                 });
         }
 
-        crawler.crawlIntervalID =
-            setInterval(
-                function() {
-                    crawler.crawl(crawler);
-                },
-                crawler.interval);
-
-        crawler.emit("crawlstart");
-        crawler.running = true;
-
-        // Now kick off the initial crawl
         process.nextTick(function() {
+            crawler.crawlIntervalID =
+                setInterval(
+                    function() {
+                        crawler.crawl(crawler);
+                    },
+                    crawler.interval);
+
+
+            // Now kick off the initial crawl
             crawler.crawl();
         });
+        crawler.running = true;
+        crawler.emit("crawlstart");
     });
 
     return crawler;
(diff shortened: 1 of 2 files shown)


bfwg:
Sorry for the multiple commits, this is my first time making a pull request, I have no idea that I should run test on my local machine before I push. I'll do that next time.

cgiffard:
Hi — thanks for filing a PR! :)

I must admit I don't quite understand the rationale behind most of these changes. Would you be able to annotate the differences in the diff? If the two are functionally equivalent, is this just a style change? And if so, what pattern did you use to decide how to make the changes?

You can fix your duplicate commits by squashing them into one! I/we generally prefer to accept PRs with clear commit messages, and no duplicate commits. [David Walsh has a handy guide here.](https://davidwalsh.name/squash-commits-git)

Christopher

bfwg:
Hi,

Ah! Somehow I included the last PR after I squashed my duplicate commits. How do I fix this. :-P
I edited my commits so that maybe it make sense now? If not please let me know. Sorry my commits are poorly written, I must do better with my English.

Fan

fredrikekelund:
Hi! So I see 3 changes going on here:

1. wait to set the main crawl interval untill the next tick. We currently wait for the next tick to make the first crawler request to enable the dev to change settings directly after having started the crawler.
2. on line 513, `resources` is used, but that array will always be empty, so we're really looking to use `list` instead.
3. in the `fetchQueueItem` method, the event listeners on `clientRequest` are registered inside of a `process.nextTick` call. According to the inline documentation, the function should return the crawler object for chaining, which it currently does not, since the return is being done inside of the `process.nextTick` call. So you removed the `process.nextTick` call there as well.

For future reference, you want to split these three changes into one commit each, and make sure to comment on each change separately so it's clear what the purpose of the change is :) Since you've already edited your commits once, you don't have to split this commit into three, though.

PR #198, opened 2015-12-03
Always match cookies with a wild card domain

intron:
Cookies instantiated without a domain get their domain property set
to "*" by default, which seems to imply that they should match any
domain, but matchDomain() did not treat them as such.

* Change matchDomain() to return true for cookies with domain of "*"

### DIFF
--- a/lib/cookies.js
+++ b/lib/cookies.js
@@ -437,6 +437,10 @@ Cookie.prototype.isExpired = function() {
 
 */
 Cookie.prototype.matchDomain = function(domain) {
+    if (this.domain === "*") {
+        return true;
+    }
+
     var reverseDomain = this.domain.split("").reverse().join(""),
         reverseDomainComp = domain.split("").reverse().join("");
 


intron:
Am I misunderstanding how this should work, or is that the expected functionality?

fredrikekelund:
Hi, thanks for the PR!

I don't actually think cookies can be domain-less, or have a wildcard domain field?

intron:
I don't think that's part of the standard, but Cookie() will set the domain property to * if one isn't provided, so I assumed the intention was to match everything. As it currently stands, matchDomain() tries to match "*" and returns false. Even though I wouldn't want my browser doing this, I can see why it might be helpful for testing things.

fredrikekelund:
Ah, silly me - I see now! In that case, this looks great :+1: If you could fix the lint error and maybe squash the 2nd commit so it only results in a single one, we can go ahead and merge this!

PR #200, opened 2015-12-10
Always match cookies with a wild card domain

intron:
Cookies instantiated without a domain get their domain property set
to "*" by default, which seems to imply that they should match any
domain, but matchDomain() did not treat them as such.

* Change matchDomain() to return true for cookies with domain of "*"


fredrikekelund:
Nice, thanks @intron!

Issue #197, opened 2015-12-02
discoverRegex need improvement


bfwg:
```js 
var crawler = new Crawler('http://www.cgiffard.com');
```
I try to test the crawler on your website :P and I got this.
`start fetching http://www.cgiffard.com/background.jpg`
`start fetching http://www.cgiffard.com/%22background.jpg%22;`
After I look into this a little bit I found out that `"background.jpg";` is passing our Regex.

I did a bit debugging and find out that 
```js
/\s?(?:href|src)\s?=\s?[^"'][^\s>]+/ig
```
Is the reason cause `"background.jpg;"` to pass.

I change it to 
```js
/\s?(?:href|src)\s?=\s?[^"'\s][^\s>]+/ig
```

Now it works like a charm.
Hope it helps.

fredrikekelund:
@bfwg could you submit this as a PR? Thanks!

fredrikekelund:
Merged fix in #203

PR #202, opened 2015-12-21
PR for #197

bfwg:


cgiffard:
Hi! Could you squash these commits?

Cheers. :)

fredrikekelund:
For documentation purposes: what we're looking to do here is to avoid picking up whitespace as part of discovered URL's.

PR #203, opened 2015-12-22
PR for #197

bfwg:
Hi, I close the old on and made a new PR.


fredrikekelund:
Looks good!

I'm just noting though that in the future, you'd want to give some detail about what the commit does directly in the commit message. This makes the repository more portable.

If you leave out the `-m` flag when committing with git in a terminal, you'll be prompted with a text editor to write your commit message. There, you write a summary in the first row, and have the ability to write a more detailed commit message on row 3 and below (row 2 needs to be empty).

Here's a good article on the subject https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message

bfwg:
I'll do that next time! Thanks for the article.

Issue #192, opened 2015-11-15
Crawl suddenly stops for no apparent reason


tedbjorling:
When I run the following code on 0.5.4 the crawler suddenly stops for no obvious reason (in the example below, I usually see that it stops when having reached approx. ~14800 queue items):

```javascript
var Crawler = require("simplecrawler");
var c = new Crawler("www.ahlens.se")
c.on("fetchcomplete", function(queueItem) {
	console.log("%d items in queue - %s",this.queue.getWithStatus('queued').length, queueItem.url);
});
c.addFetchCondition(function(obj) {
	if ((obj.path.match(/\.css/i) != null)
	 || (obj.path.match(/\.ico/i) != null)
	 || (obj.path.match(/\.jpg$/i) != null)
	 || (obj.path.match(/\.gif$/i) != null)
	 || (obj.path.match(/\.png$/i) != null)
	) {
		return false;
	}
	return true;
});
c.start();
```

If I freeze/defrost the queue and resume the crawl session, it continues for essentially the same duration after which it stops again, and so on.

cgiffard:
So it fires `fetchcomplete`, even where the queue has a considerable number of unfetched items? :-/

tedbjorling:
I guess the `fetchcomplete` events are expected when resources have been fetched and are ready for processing. The `complete` is never emitted, so basically it just seems to be left hanging there. Are you able to reproduce this?

nttdocomo:
I have the same problem

tedbjorling:
After having done some debuging I made two fixes that appears to have remedied this for me at least. I realize I could have been a bit more methodical, but fwiw, the following greased the gears for me:

1) I was using an older version of Node.js (0.10.33), upgrading to 5.1.1 made things run much smoother.

2) I saw issues with the "receiveData" function in crawler.js, where it sometimes did not enter the code block calling processReceivedData(). Reason being that the http(s) response did return an amount of data less than the states "content-length", meaning that things essentially came to an halt. If I send through wether the http(s) response is "data" or "end" to  the "receiveData" function, and ensure that the code block calling processReceivedData() is invoked if the http(s) response has come to an end, I experience no more issues of the kind stated above.

fredrikekelund:
@tedbjorling could you submit your changes in a PR? Sounds like this could do some good if merged

fredrikekelund:
I've tried to address this in #209, would be great with some additional testing of that fork! Ping @tedbjorling

[...]

PR #209, opened 2016-02-03
New behavior for closing requests

fredrikekelund:
This is an attempt at addressing #192. I've also experienced the problem where certain requests are just "left hanging", they're never closed and cause the process to keep running indefinitely unless timed out.

My solution here was to always close HTTP responses on their "end" event.

I don't see any obvious holes in the old logic, my best guess is that the `complete` property for some reason isn't always set on the response when it is in fact complete. I don't see why that would be, but on the other hand I couldn't find any documentation for that property, so there might be some unpredictable behavior there.

The diff is noisier than my actual changes because of an indentation change, but the basic changes are:
1. call `processReceivedData` directly on the request's "end" event
2. always assume that the `chunk` argument to `receiveData` is defined
3. remove the logic in `receiveData` that calls `processReceivedData`

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -1050,9 +1050,12 @@ Crawler.prototype.handleResponse = function(queueItem, response, timeCommenced) 
 
     // Function for dealing with 200 responses
     function processReceivedData() {
-        if (queueItem.fetched) {
+        if (dataReceived || queueItem.fetched) {
             return;
         }
+
+        responseBuffer = responseBuffer.slice(0, responseLengthReceived);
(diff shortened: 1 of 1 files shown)


cgiffard:
Thanks! I'll check this out when I get to work today. :)

fredrikekelund:
If the `complete` property is the actual culprit, then the [response.finished](https://nodejs.org/api/http.html#http_response_finished) might be what we're really looking for. However, I still don't see a reason to not use this new behavior, it's also clearer from a developer's perspective

Issue #153, opened 2015-07-27
Add support for robots.txt, on by default


mvl22:
We are seeing sites receiving crawls with this bot's signature which disobey robots.txt, falling into our honeytrap.

Please kindly add support for robots.txt, (strongly) preferably on by default.

I would argue that crawlers have a responsibility to ensure that they act as a well-behaved web citizen by default. This applies even more to open-source code crawlers, which are likely to be used by people the author has no control over.

So please also kindly ensure this behaviour is on by default. If someone is installing this crawler, and they are using it on their own website, they will have full knowledge of why robots.txt rules are in place, and so can decide to disable the feature. If they are using it on another site, then the robots.txt should be respected as a matter of principle.

>  I figure that because it's a specialised tool, and not designed to crawl websites indiscriminately [1]

As open source code, it absolutely can be used in that way, I'm afraid.

[1] https://github.com/cgiffard/node-simplecrawler/issues/59#issuecomment-25943001

unr3al011:
+1

mortenf:
+1

marcusgreenwood:
+1

lgraubner:
+1

davidgljay:
+1

fredrikekelund:
I've used the [robots-parser](https://www.npmjs.com/package/robots-parser) module to successfully add support for respecting robots.txt rules. And I actually think this is outside the scope of simplecrawler, better to focus on the basics and expose API's to easily add this kind of behavior when needed.

Here's a code example of how to add support for robots.txt rules:

```js
var Promise = require("bluebird"),
	request = Promise.promisify(require("request")),
	parseRobots = require("robots-parser");

var robotsUrl = "http://example.com/robots.txt";

request(robotsUrl).spread(function (res, body) {
	var robots = parseRobots(robotsUrl, body);

	scraper.crawler.addFetchCondition(function (parsedUrl) {
		var standardUrl = url.format({
			protocol: parsedUrl.protocol,
			host: parsedUrl.host,
			pathname: parsedUrl.path.split("?")[0],
			search: parsedUrl.path.split("?")[1]
		});

		var allowed = false;

		// The punycode module sometimes chokes on really weird domain
		// names. Catching those errors to prevent scraper from crashing
		try {
			allowed = robots.isAllowed(standardUrl, userAgent);
		} catch (error) {
			logger.error("Caught error from robots.isAllowed method on url %s", standardUrl, error);
		}

[...]

PR #212, opened 2016-02-05
Automatic robots.txt parsing

fredrikekelund:
As discussed in #153, this PR brings on-by-default support for robots.txt parsing.


fredrikekelund:
The code changes I've made are:
- added a `Crawler.prototype.getRobotsTxt` method, which does the actual retrieving of the robots.txt file
- abstracted logic from `Crawler.prototype.fetchQueueItem` into `Crawler.prototype.getRequestOptions` so that it can be reused in `getRobotsTxt`
- added robots-parser module and a fetch condition that's stored outside of the `_fetchConditions` array for determining whether a URL is to be fetched or not

lib/crawler.js
```diff
@@ -230,42 +237,222 @@ util.inherits(Crawler, EventEmitter);
...
+            crawler._robotsTxtFetchCondition = function (parsedURL) {
+                parsedURL.hostname = parsedURL.host;
+
+                var formattedURL = uri(parsedURL).href(),
+                    allowed = false;
+
+                // The punycode module sometimes chokes on really weird domain
+                // names. Catching those errors to prevent crawler from crashing
+                try {
+                    allowed = robots.isAllowed(formattedURL, crawler.userAgent);
```
cgiffard: Do we want to emit a `fetcherror` event or similar? I figure we probably shouldn't silently swallow the URLs.

lib/crawler.js
```diff
@@ -230,42 +237,222 @@ util.inherits(Crawler, EventEmitter);
...
+
+            requestOptions.headers[header] = crawler.customHeaders[header];
+        }
+    }
+
+    return requestOptions;
+};
+
+/*
+    Public: Initiates a request for the robots.txt resource on crawler's target
```
cgiffard: I'm wondering whether we need a cached list of hostnames, with a parsed robots-txt for each one — if the crawler, using this patch, continues to another hostname, then it won't respect robots.txt over there.

(I'd consider merging this but adding that as a further issue to be addressed.)

lib/crawler.js
```diff
@@ -230,42 +237,222 @@ util.inherits(Crawler, EventEmitter);
 Crawler.prototype.start = function() {
     var crawler = this;
 
-    // only if we haven't already got stuff in our queue...
-    crawler.queue.getLength(function(err, length) {
-        if (err) {
-            throw err;
-        }
+    crawler.getRobotsTxt(function (error, robotsTxtUrl, robotsTxtBody) {
```
cgiffard: Error should be an error object, methinks (or if you wanna use a string, build an event and emit it instead.)

lib/crawler.js
```diff
@@ -230,42 +237,222 @@ util.inherits(Crawler, EventEmitter);
...
+                responseBuffer = new Buffer(responseLength),
+                responseLengthReceived = 0;
+
+            response.on("data", function (chunk) {
+                if (responseLengthReceived + chunk.length <= crawler.maxResourceSize) {
+                    chunk.copy(responseBuffer, responseLengthReceived, 0, chunk.length);
+                    responseLengthReceived += chunk.length;
+                } else {
+                    response.socket.destroy();
+                    callback("resourcetoobig");
```
cgiffard: This should really be an error object with a nice message. Or emit it. :)

lib/crawler.js
```diff
@@ -230,42 +237,222 @@ util.inherits(Crawler, EventEmitter);
...
+    var requestOptions = crawler.getRequestOptions({
+        protocol: crawler.initialProtocol,
+        host: crawler.host,
+        port: crawler.initialPort,
+        path: "/robots.txt",
+        url: robotsTxtUrl
+    });
+
+    // Get the resource!
+    var clientRequest = client.request(requestOptions, function(response) {
```
cgiffard: At this point I'm wondering whether abstracting the request logic might be useful. I feel like we're repeating ourselves a lot here.

lib/crawler.js
```diff
@@ -751,12 +938,11 @@ Crawler.prototype.queueURL = function(url, queueItem) {
...
 
     // Pass this URL past fetch conditions to ensure the user thinks it's valid
-    var fetchDenied = false;
-    fetchDenied = crawler._fetchConditions.reduce(function(prev, callback) {
+    var fetchDenied = crawler._fetchConditions.reduce(function(prev, callback) {
         return prev || !callback(parsedURL);
     }, false);
 
-    if (fetchDenied) {
+    if (fetchDenied || crawler._robotsTxtFetchCondition(parsedURL)) {
```
cgiffard: I don't really understand why there's a private function for this — I'm thinking we make it public and stick it directly on the prototype. Then if there's no robots.txt, or the checking has been disabled, we make that an initial check inside the body of the function and simply return true.

lib/crawler.js
```diff
@@ -832,86 +1017,14 @@ Crawler.prototype.fetchQueueItem = function(queueItem) {
...
-        }
-    }
-
     // Apply the ignoreInvalidSSL setting to https connections
     if (client === https && crawler.ignoreInvalidSSL === true) {
         client.rejectUnauthorized = false;
         client.strictSSL = false;
     }
 
+    var requestOptions = crawler.getRequestOptions(queueItem);
```
cgiffard: I like it. :)

cgiffard:
Looks really good for a first cut. I think we can smooth it out a lot, but this is awesome! Thanks so much. :)

lib/crawler.js
fredrikekelund: Fair point, I'll pass an actual Error and some more verbose error messages!

lib/crawler.js
fredrikekelund: Yeah, I know - I think the main culprit here is the fact that the `getRequestOptions` logic expects queueItem.url to be defined as well as all the individual parts of the URL. Maybe I could change it so that if only a URL is provided to getRequestOptions, it knows how to construct a `http.request` options object from just that. That helps us avoid some repetition here

lib/crawler.js
fredrikekelund: My thought here was that by not placing this in the fetch conditions array, it would be less prone to being accidentally removed by the developer, but I don't know how much of a concern that is. But it would be more elegant to place it in the fetch conditions array, I agree. Which do you prefer?

lib/crawler.js
fredrikekelund: Oh, that's right - I didn't think of the multiple hostnames use case... Maybe it's a good idea to flesh that out in a separate PR, but I guess that would probably mean that getRobotsTxt should be called as part of fetchQueueItem instead, when the hostname isn't in the hostname cache. 

That might add some extra complexity around the robots.isAllowed call where we need to check what hostname the current fetch condition concerns itself with. But that might also work out of the box because robots-parser actually does know the URL of each parsed robots.txt. I'm not sure which right off the bat.

Come to think of it, maybe I'll give it a go to see if it's easy enough to implement as part of this!

lib/crawler.js
fredrikekelund: Good point, how do you feel about emitting a new event, maybe calling it `fetchdisallowed`?

fredrikekelund:
Thanks! Yeah, I'll take a stab at fixing the stuff in the line notes and we can hopefully make it come out great

lib/crawler.js
cgiffard: Well it's a "condition by which we fetch things", but it's more than a fetch condition, I felt. So I wouldn't necessarily expose it as a fetch condition, I'd expose it as a top level method similar to the other methods to parse and validate URLs.

fredrikekelund:
@cgiffard I've comitted some stuff that addresses some of the issues we discussed, have a look and let me know what you think!

I should probably add some tests for the two new events: "robotstxterror" and "fetchdisallowed" as well

test/testcrawl.js
```diff
@@ -47,6 +47,41 @@ describe("Test Crawl", function() {
...
+            queueItem.url.should.not.equal("http://127.0.0.1:3000/forbidden");
+        });
+        crawler.on("complete", function() {
+            done();
+        });
+    });
+
+    it("should be able to disregard rules in robots.txt", function(done) {
+
+        this.slow("2s");
```
cgiffard: Is it taking two seconds? (I'm just worried that if we get too many of these slow tests they'll all compound to create a test suite that takes ten years to run, haha.)

fredrikekelund: Takes between 1500 and 1800 ms on my machine. I could probably speed this up by listening to the new `fetchdisallowed` event instead, though

fredrikekelund:
I've updated the tests, they run a lot quicker now. The total test run time was 7s before my last commit, now it seems to be about 4s.

fredrikekelund:
I've added logic that should deal with multiple hostnames (protocol and port changes count as well). Now we just need to add tests for it. I'll make it so it starts a second instance of testserver on a different port. Thoughts on how to deal with the tests in the best way, @cgiffard?

cgiffard:
Thanks! I'll have to give it a good look over.

I'm out all tomorrow but should be back my Sunday (probably your late Saturday) to get some work done here. :)

fredrikekelund:
Cool! I'll see if I have time to expand on the tests in the meantime

fredrikekelund:
@cgiffard I changed the way the test server is constructed to allow for more than one to run simultaneously. So now, I kick off a second one, with a second (small) set of routes, to ensure that the crawler does in fact keep track of different robots.txt's for different hosts.

I feel pretty good about the state of this PR now, let me know what you think and hopefully we can merge this pretty soon :)

cgiffard:
I must admit I haven't had a great deal of time to have a look at this this weekend — I'm really sorry! I'll get on it ASAP, but given the scope of the change I'd like to have a good look over before merging. :)

Apologies again! :(

fredrikekelund:
That's alright, but it should be stable enough now for a final judgment now anyway. I definitely understand that you'd like to review it properly. Let me know when you've had a look :)

VFe:
I played with the latest commit on this fork, I tried it on several sites and it didn't honor the robots.txt. 

Though another (minor) issue is that some legit robots.txt will come from a 301(when the crawler tries to fetch from my einsfeld.com domain it fails due to the 301). 

Node 5.6.0/OSX 10.11

fredrikekelund:
Good point, I guess we should follow redirects as well. I'll try it some more on live sites later today, do you have any examples of which sites it didn't work with when you tried, @vfe? Thanks!

fredrikekelund:
@VFe I realized that response bodies could come out as total gibberish since I didn't account for the possibility of the response being gzip encoded. I still haven't fixed the redirect problem, but would you care to run another test on the sites you were having trouble with?

fredrikekelund:
And there's the redirect support as well!

Might be worth noting that there's no maximum number of allowed redirects right now, it just keeps on following the redirects. I don't think this is much of a problem, but is you think we should account for it, @cgiffard, I can add it in.

cgiffard:
I feel like this is starting to get really complex, and it might be worth evaluating whether simplecrawler would be better off for some architectural changes to support something like this.

Creating a robust API which would more easily allow resources to be fetched and excluded would make a lot of this much easier, as well as adding key injection points in the crawl lifecycle.

cgiffard:
(Not saying we shouldn't merge this feature, but that we might want to schedule a big refactor for a milestone release.)

cgiffard:
I wonder whether a simpler way of realising this feature might be to automatically push a `{domain}/robots.txt` into the queue whenever a new domain is discovered — and handle that response specifically by parsing it and creating a fetch condition with the result. That way we don't have to maintain custom request logic just for the `robots.txt` file, and can support an indefinite number of domains.

cgiffard:
(I wish I'd thought of that idea a really long time ago — sorry @fredrikekelund !!)

fredrikekelund:
@cgiffard you're right, that is a simpler way of realising this, but somewhat less clean IMO. `robots.txt` probably wouldn't ever be discovered, so us adding it manually to the queue and treating the response differently would skew the queue length, and we might not want to emit "fetchcomplete" events etc for it.

And even though the diff is big here, the biggest addition that wouldn't be there if we went with the other way of implementing this is the getRobotsTxt method, which isn't *huge*. With the other implementation, we'd still have the fetch conditions and touchedDomains list etc.

Speaking of the architectural changes: we have merged a breaking change (fetch404 event stuff), so I'm thinking that we ought to land this change and perhaps the "download conditions" that I've been going on about (which would also mean dropping `fetchWhitelistedMimetypesBelowMaxDepth`) and then release a 1.0.0.

1.0.0 is the big number of course, but I think there's more value in simply adhering to semver and trying to move forward at a reasonable pace than us waiting to release 1.0.0 until we've made all architectural changes that we want, better to make that 2.0.0.

VFe:
domains it had failed on previous were einsfeld.com, sportarm.com, and http://www.jecauthen.com(note, I own the first one, but not the other two), with einsfeld /images/ is forbidden but fetched, for the other two I changed the crawler.initialPath to be something forbidden, and they'd still get fetched.

I'll test out changes later today when I have a chance.

VFe:
Tried out the changes, the re-direct issue seems fixed, it manages to retrieve the robots.txt from my einsfeld.com domain. But it still doesn't honor the disallows and fetches images.

Been digging around trying to see if I can find what's going on, but I'm fairly new to both node and this library, and haven't found anything helpful.

fredrikekelund:
Thanks again for taking the time to test out this branch, @VFe! The fault was with the URL construction in the `urlIsAllowed` method. That's been fixed now, and the robots.txt condition worked as expected for einsfeld.com when I tried it out, feel free to give it a try yourself!

VFe:
I tested it against several sites and it works like a charm!

The only issue I see now is the initialPath issue I mentioned above, if I set crawler.initialPath = /forbidden path/ it'll still fetch it as the first item even though it's forbidden.

That said, that's an edge case of me going out of my way to break it :), so I think it's pretty solid as is with just the caveat of knowing that.

Great work! Glad I could help a little.

fredrikekelund:
I'll double check the problem with initialPath, it's supposed to avoid the first URL as well, if disallowed. I noticed that your domain redirects to the www subdomain, is it possible that the first item is fetched, but only from the non-www domain?

fredrikekelund:
The problem was that the queueItem wasn't being given another status when disallowed from the get go, which meant that it would be fetched in the next crawl tick (as robots.txt filtering happens when discovering URL's, not before fetching them).

VFe:
Ran into another issue that stops the crawler, after crawling for like 30min~ on this one particular website with a lot of issues, I'll get this error: 
screen shot 2016-03-01 at 8 20 08 pm

It seems like the crawler is periodically re-fetching the robots.txt file? and if it fails to fetch enough times it'll spit this out. Website is connecticutshotgun.com(I'm not the owner), where I can consistently re-create this.

fredrikekelund:
Thanks again for testing this fork so diligently @VFe! I ran some tests on the same site and discovered that the problem was in how I constructed the URL that we put in the cache of domains that we've retrieved a robots.txt for. All URL's with a query string were considered unique hosts, and that caused some funky `robots.txt?add_to_wishlist=1726` like URL's and weird behavior.

AFAICT this fixes the problem, would be nice if you could confirm that it works for you as well!

VFe:
Been running this for awhile now, no new problems to report related to the robots functionality. Seems good :)

fredrikekelund:
Thanks for the report, @VFe! I'm inclined to merge this pretty soon unless @cgiffard has any considerations?

cgiffard:
I reckon we should go ahead for merge, if you're happy @fredrikekelund. I kinda wanna revisit the logic along the lines of my previous comment, but might file a PR to demonstrate my idea later on. Better to keep things rolling. :)

Thanks again for your help, and sorry for the delay!

fredrikekelund:
@cgiffard cool, thanks for the go ahead! I'll have time to do some simplecrawler work this friday, will merge this then

PR #216, opened 2016-02-13
Find resources in <meta> redirects

### LINKED ISSUE #190: Crawler ignores "HTML redirects"
Example:

     <meta http-equiv="refresh" content="0; url=en/index.html">

My crawler ignores that, and I don't find any code to handle it. I'm going to have to extend the discovery function (again).

fredrikekelund:
We only discover those resources, we don't treat them as an actual
redirect (eg. emit a "fetchredirect" event). Fixes #190

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -169,7 +169,20 @@ var Crawler = function(host, initialPath, initialPort, interval) {
         // strings out of javacript: URLs. They're often popup-image
         // or preview windows, which would otherwise be unavailable to us.
         // Worst case scenario is we make some junky requests.
-        /^javascript\:[a-z0-9\$\_\.]+\(['"][^'"\s]+/ig
+        /^javascript\:[a-z0-9\$\_\.]+\(['"][^'"\s]+/ig,
+
+        // Find resources in <meta> redirects. We need to wrap these RegExp's in
+        // functions because we only want to return the first capture group, not
+        // the entire match. And we need two RegExp's because the necessary
+        // attributes on the <meta> tag can appear in any order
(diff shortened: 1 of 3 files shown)


lib/crawler.js
```diff
@@ -559,9 +563,17 @@ Crawler.prototype.discoverResources = function(resourceData, queueItem) {
     // Rough scan for URLs
     return crawler.discoverRegex
         .reduce(function(list, regex) {
+            var matches = resourceText.match(regex);
+
+            // If the RegExp doesn't contain a global flag, we expect to find the
+            // resources in the capture groups, so we strip away the match as a
+            // whole (ie. the first entry in the matches array).
+            if (regex.toString().split("/").slice(-1)[0].indexOf("g") === -1) {
```
cgiffard: Whoah, this is meta. I wonder if there's a simpler way to do this.

fredrikekelund: To avoid the entire split and slice part, there's https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/flags, but that's not supported in node.

But I just realized that to avoid this entire logic (which I admit might not be the most solid) I could actually just wrap the entire regexp in a non-capturing group! I'll fix that later today

fredrikekelund:
I went for a slightly more generalized solution: allowing functions in the discoverRegex array. What do you think about that, @cgiffard?

cgiffard:
Looks good. In future I think a good philosophy might be to make a simpler but more general regex, which might hit false positives, but which should be validated and filtered out later. :)

Issue #114, opened 2015-03-19
Implementing a link checker


davidsinclair:
Hi,

First off I have to say I really like node-simplecrawler - good work! 

I want to implement a link checker for the site I am working on and using it has gotten me 95% of the way there but I am stuck getting the last little bit working. What I want to do is check all links on the site for 404's and redirects.  We feed the results into our custom content management system so that content authors can fix the broken links it finds.

However, we want to verify external links but not to actually go an and crawl that site.  It seems that simplecrawler will not fetch the external links unless I set filterByDomain to false.  But if I do that then it starts to also crawl those sites.  I do not want to set a fixed depth because I want to ensure our whole site gets crawled.  

I have tried implementing my own link discovery but when I do this the queueItems do not have a referrer set.  I really need to know the referrer in order to report the link back to our system.

Any ideas?  For now I have just added an if statement to your discoverResources function that returns an empty array for external sites.  Is there a better way for me to do this? 

Many thanks for any guidance you can give me,

David

cgiffard:
Hi! This is a good use case — thanks for filing an issue. I'm pretty busy for a couple of days, so I'll have a think about this and get back to you on the weekend. :)

davidsinclair:
Hi,

Have you had a chance to think about how this might be handled?  I would be happy to implement the solution and submit a pull request if you could suggest the approach you think fits best.  

Thanks,

David

cgiffard:
Hey — if you're still interested, the feature is being bashed out in PR #74 (https://github.com/cgiffard/node-simplecrawler/pull/74) — but I've asked for a test to be added. I'll pick it up myself if required — seems a lot of people want this feature. :)

fredrikekelund:
This ticket is an oldie, but I guess what could solve this use case would be to pass the current queueItem to fetch conditions. With that, I could tell where the resource I'm looking at has been discovered, and filter it based on that. So I set filterByDomain to false, but avoid any resources on the second site that were in fact discovered on the second site.

fredrikekelund:
The feature I suggested was merged in #217 and released in 0.6.2

Issue #131, opened 2015-05-01
Ignoring Max Depth


BenChirlin:
Hi. I was wondering if there's a way to have a set max depth but then have that ignored for certain types of URLs. I saw the `fetchWhitelistedMimeTypesBelowMaxDepth` but the documentation seems to imply this is only applicable to assets and not actual URLs in the page.

Specifically I want to crawl my site from an origin request to a given depth but want to ignore that depth with pagination links on pages to ensure the spider will crawl an entire paginated post and not cut off after a certain number of pages due to `maxDepth`. 

For instance, if `spider.maxDepth = 3;` and the spider is on a page `www.mysite.com/2014/06/astory.html`, a depth 3 from it's origin request. Is there a way to ensure the spider will continue to crawl a link `www.mysite.com/2014/06/astory.html?pg=2` even though it's already at its `maxDepth`?

Thank you!

BenChirlin:
After a quick dive into the source it looks like the [`fetchWhitelistedMimeTypesBelowMaxDepth`](https://github.com/cgiffard/node-simplecrawler/blob/master/lib/crawler.js#L344) property only checks an internal list of type options. I'll make a pull request with that list being a variable which defaults to the current configuration.

fredrikekelund:
The `fetchWhitelistedMimeTypesBelowMaxDepth` setting is a bit weird to me... I think the best way to accomplish what you're after would be to provide the current queueItem to all fetch conditions. That would enable the developer to determine whether to fetch the next resource based on the current depth.

I realize this ticket hasn't been active in a while, but I'll see what I can do about submitting a PR for this!

BenChirlin:
Yeah pretty old. I already finished my project but I ended up forking the repo, adding an `ignoreDepth` option for a specific regex, and managed it that way. I submitted a PR but the manager of the repo said he didn't feel like it added enough utility but I'm glad to have tried to contribute (and got my project finished!).

cgiffard:
I like the idea of making the queueItem available to fetch conditions. Bear in mind that it won't have much data in it at that early stage. (I think in the life cycle of where the fetch condition is executed, the queueItem hasn't even been created yet.)

cgiffard:
Generally speaking I think a lot of people's requests centre around specific rules for what is fetched and what isn't — and having a more flexible mechanism to allow them to write those rules would both satisfy their requirements as well as keep a lot of crazily complex and highly specific logic out of simplecrawler core.

[...]

PR #217, opened 2016-02-13
Pass queueItem to fetch conditions

fredrikekelund:
This enables more sophisticated logic in fetch conditions. It should also be just what the doctor ordered in #114.

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -753,7 +753,7 @@ Crawler.prototype.queueURL = function(url, queueItem) {
     // Pass this URL past fetch conditions to ensure the user thinks it's valid
     var fetchDenied = false;
     fetchDenied = crawler._fetchConditions.reduce(function(prev, callback) {
-        return prev || !callback(parsedURL);
+        return prev || !callback(parsedURL, queueItem);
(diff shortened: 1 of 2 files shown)


lib/crawler.js
```diff
@@ -753,7 +753,7 @@ Crawler.prototype.queueURL = function(url, queueItem) {
     // Pass this URL past fetch conditions to ensure the user thinks it's valid
     var fetchDenied = false;
     fetchDenied = crawler._fetchConditions.reduce(function(prev, callback) {
-        return prev || !callback(parsedURL);
+        return prev || !callback(parsedURL, queueItem);
```
cgiffard: Wow, didn't think that'd be so simple! Nice catch. :)

fredrikekelund:
Yeah, was pretty happy when I realized :) What do you think about my idea in #131 btw? I think "download conditions" could be a second concept that gives greater flexibility around which resources to fetch, and which to avoid. We could drop `fetchWhitelistedMimeTypesBelowMaxDepth` as well, which IMHO seems like a setting that really is fairly unique for certain use cases

PR #223, opened 2016-02-15
Use initial protocol and port in `processURL`

Hywan:


cgiffard:
Hey! Thanks for submitting a pull request. So we can get a better idea of what's going on (simple as the change appears) — would you mind providing a bit of explanatory information as to your rationale here? Also, if this is fixing a bug, would you mind including a regression test for the feature?

Thanks! :)

Hywan:
To be frank, I have a bug with the crawler and this patch does not fix it, but I was thinking this is weird to have a initial port and an initial path and not using it. Why hardcoding the default port and path in this particular case :-)?

cgiffard:
Hey — thanks for getting back to me. :)

[If you look at the code directly above your patch](https://github.com/Hywan/node-simplecrawler/blob/patch-1/lib/crawler.js#L424-L425) we're actually using URIjs to normalise the URL, so the port will be set relative to the absolute URL of the page from which it was discovered, unless it sets an absolute hostname or port. I am fairly sure that `port()` and `protocol()` will never be falsy due to this, so the code you've added won't ever run.

Could you talk about the problem you're having? Perhaps file a separate issue for it?

Hywan:
However, if we have `https://foobar.com`, the default port will be 80, which is wrong :-/. Is it an known issue?

cgiffard:
Is this a problem you've experienced? My understanding is that URIjs will retrieve the appropriate port based on the protocol.

dzafel:
@cgiffard
My test code (`URL` and `context.uri` based on real data I had when crawling a website):
```
newURL = uri('https://www.example.com/my-portfolio')
  .absoluteTo('https://www.example.com:443/')
  .normalize();
```

result:
```
{
  protocol: 'https',
  username: null,
  password: null,
  hostname: 'www.example.com',
  urn: null,
  port: null,
  path: '/my-portfolio',
  query: null,
  fragment: null
}
```
so as you can see `port` is empty, but it should be 443. Tested with urijs 0.16, 0.17 and 0.18

cgiffard:
Does it cause a bug — are you unable to connect using those details? My understanding is that the use of the https agent should default to 443, but I could be wrong. (I'll have to review the code to check, in a bit.)

dzafel:
From my experience it's not possible to connect to addresses like `https://example.com:80/my-portfolio` when a website is on Windows Server (IIS) - because the `:80` part is added by simplecrawler when `newUrl.port()` is empty

cgiffard:
Ah, I see. It would be good to get a unit test in for this behaviour. I'm really flat out at the day job and unable to touch this in any meaningful way until the weekend — but I'll have a look then!

Issue #220, opened 2016-02-14
Different events for 404 and 410 


tmpfs:
I think the event name ambiguity is a little misleading (although it is documented in the readme that 410 triggers `fetch404`). I have implemented a quick fix for this to distinguish the event names if you want the pull request?

My rationale is that `fetch404` should not have the `410` event status code bundled in because there are instances when they would need to be handled differently.

My use case is that I am building a tool that downloads all `text/html` and `application/xhtml+xml` pages and validates them using the [nu validator](https://github.com/validator/validator). I want to add the logic to download and validate 404 errors but ignore 410 errors. Clearly I can just switch on the status code in `fetch404` but there is something about the event name that bothers me.

I suggest that either:

* `fetch404` and `fetch410` are emitted (quickly implemented, tests passing)
* the event name is renamed to `notFound` or something similar

Sorry if this is a little pedantic.

tmpfs:
Clearly, renaming to `notFound` is a breaking change and requires a minor version bump, as does distinguishing event names too (people might already be switching on status code in `fetch404`).

cgiffard:
Good suggestion, and definitely an artefact of simplecrawler's evolution. I'm thinking we could add a `notFound` event which would take both 404s and 410s, but remove the 410 responses from emitting a `fetch404`, which is clearly confusing.

tmpfs:
I was thinking it wouldn't hurt to emit `notFound`, `fetch404` and for consistency also emit `fetch410`. That should cover all requirements. I don't think emitting the additional events would impact performance much, this should do the trick:

```javascript
// Emit 404 event
crawler.emit("fetch" + response.statusCode, queueItem, response);
crawler.emit('notFound', queueItem, response);
```

fredrikekelund:
I think separating the 404's and the 410's sounds like a sound idea! I'm not 100% sure that making 3 events out of 1 feels like the best way to go about this though (but I'm not 100% sure it's the wrong way either). Instinctively I'd probably vote for doing one `fetchnotfound` (404 and 410) and one `fetch404` (just 404).

tmpfs:
You are correct, three is not the way to go, thinking some more, my preference is `fetch404` and `fetch410`, then when both need to be handled with the same logic:

```javascript
crawler.on('fetch404', onError)
  .on('fetch410', onError);
```

Done :)

[...]

PR #224, opened 2016-02-16
Separate 404 and 410 events.

tmpfs:
* The 410 status code triggers a fetch410 event.
* The 404 status code triggers a fetch404 event.

Tests updated and passing, this is a breaking change and requires a
minor version bump.

See #220 for more information.

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -1211,7 +1211,7 @@ Crawler.prototype.handleResponse = function(queueItem, response, timeCommenced) 
         queueItem.status = "notfound";
 
         // Emit 404 event
-        crawler.emit("fetch404", queueItem, response);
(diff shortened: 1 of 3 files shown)


fredrikekelund:
Looks great! Could you add a second test for the fetch404 event as well? Figure we might add a test for the other event if we split the fetch404 event into two

tmpfs:
Oops, I assumed that test was in there - my bad, added another commit which tests `fetch404`.

fredrikekelund:
Thanks! As you say, this is a breaking change, so I actually think it should be one of the inclusions in a 1.0.0 release. I'll try and submit a few other changes over the course of the coming week to make sure we aggregate a good set of changes for 1.0.0

Issue #230, opened 2016-02-22
Question: should fetch conditions continue processing when fetch is denied?


tmpfs:
At the moment if appears the current fetch condition logic continues to process fetch conditions even when a fetch has been denied by a previous fetch condition.

I have two fetch conditions, the first is denying the fetch, the second is allowing it, shouldn't the fact that the first fetch condition denied it prevent the second fetch condition from being called?

For me this makes support for multiple fetch conditions somewhat unusable as my fetch conditions are doing completely different tasks. Obviously I can just bundle all logic into one fetch condition but then that would make the code very unwieldy :(

What do you think?

tmpfs:
I think the culprit is here:

https://github.com/cgiffard/node-simplecrawler/blob/59c42fac674ce548d37bf916b6b20b125ed2d03b/lib/crawler.js#L770-L773

When I simplify it to:

```
    var fetchDenied = false;
    for (var i = 0; i < crawler._fetchConditions.length; i++) {
        fetchDenied = crawler._fetchConditions[i](parsedURL, queueItem);
        if (fetchDenied === false) {
            break;
        }
    }
```

Then the behaviour is as I would expect.

All the current tests pass with this, would you accept a PR for this?

cgiffard:
Hey, thanks for filing an issue and your PR. 

Simplecrawler's strategy is fetch by default and iteratively exclude — and so fetch conditions are exclusive. If you've got five hundred, and one returns false, the resource will not be fetched.

I've always seen early bailout as more of an optimisation issue (and most people don't seen to have particularly complex fetch conditions) than a UX one, so I'm curious to hear about why this was causing a problem for you.

Your proposed solution might very well be the right one, but it helps to have context. :)

cgiffard:
It also appears that the code /should/ be working for early bailout. Could you provide a test case where it isn't?

tmpfs:
Sure, I understand.

Well my first fetch condition was so solve the strict cross-domain issue (hopping different ports on the same domain) then I wanted to allow users to filter using a `-p | --pattern` regexp pattern, so I implemented the code and tested it thoroughly.

When I went to add that extra fetch condition my `--pattern` filter fetch condition was returning false (correctly) and the second fetch condition was returning true (correctly) hence I figured this was an issue with the fetch condition logic.

By quickly switching the second condition to return `false` my pattern matching fetch condition worked as expected.

tmpfs:
Ha, you are correct - again! :)

[...]

PR #231, opened 2016-02-22
Patch fetchDenied logic.

tmpfs:
So that a fetch condition that returns false prevents any further fetch
conditions from being called.

Note that this code can be simplified further by just calling `return` rather than `break`, I'll update this PR if you think this logic is correct and want to merge.


fredrikekelund:
Thanks for bringing this up! As you concluded in #230, if just one fetch condition returns false, the resource will be avoided. And even though we iterate through each of the remaining entries in the `fetchConditions` array, they aren't actually executed (if a previous one has returned false), so I don't think there's much to gain performance wise here :/

tmpfs:
My bad for not doing the isolation test first, apologies!

Issue #232, opened 2016-02-22
Strange javascript resources


tmpfs:
Just logging this before I filter them out from my crawl:

```
 WARN | 404 http://localhost:3000/assets/js/vendor/%29.revokeObjectURL%28file
 WARN | 404 http://localhost:3000/assets/js/vendor/blob
 WARN | 404 http://localhost:3000/assets/js/vendor/%29.createObjectURL%28blob
```

Generated by this file:

https://github.com/eligrey/FileSaver.js/blob/master/FileSaver.js

It appears to me that possibly this is an error with the discover resources logic for javascript files.

fredrikekelund:
@cgiffard's philosophy has been to keep simplecrawler's discovery mechanism "dumb", with the reasoning that it's better for the crawler to make a few junk requests than for it to miss some stuff. So by default, the `discoverResources` method will pick up some garbage, the idea is to make it easy to replace this logic with more a more advanced discovery function, tailored for the dev's specific needs.

If there is a specific, non-huge improvement to one of the default discoverRegex's, feel free to submit a PR, otherwise, maybe the way to go in your case would be to override the discoverResources method? When I've worked with simplecrawler, I've had a lot of success using cheerio to parse pages and do a more refined extraction of resources!

tmpfs:
Thanks, I just got a feel for that from reading the `base href` issue.

Now I have my pattern matching support in place, it's pretty easy for me to ignore these sorts of things without implementing `discoverResources`.

Cheers for the tip on `cheerio`, looks quite nice :)

Closing as expected behaviour.

tmpfs:
For the record it looks like these patterns:

```
        /\s?url\((["']).*?\1\)/ig,
        /\s?url\([^"'].*?\)/ig,
```

For CSS I imagine, I think we could remove the `i` flag and be safe, it then wouldn't match the junk in this issue and I don't know of anyone that uppercases `url()` in CSS files and likely uppercase is invalid CSS anyhow (I would have to check).

What do you think?

tmpfs:
Spec is quite specific:

https://www.w3.org/TR/CSS2/syndata.html#uri

I know css selectors are case-insensitive but I think we could be safe to remove those `i` flags, @cgiffard - what do you think?

tmpfs:
Just confirmed that removing the `i` flag does indeed prevent those junk requests.

cgiffard:
[...]

PR #233, opened 2016-02-22
Remove case-insensitive flags from css regexp.

tmpfs:
I think this is ok, obviously feel free to correct me! :)


fredrikekelund:
Closing this in accordance with #232

PR #239, opened 2016-03-03
supports srcset, closes #237

### LINKED ISSUE #237: Support srcset
Since `<picture>` elements use `srcset`, and they are a chore to check and debug, the checking of `srcset` attributes would be helpful.
I think we would only have to improve the `crawler.discoverRegex` to check for `srcset` as well and modify the regex to stop at a whitespace or quote/doublequote, since a `sourceset` can look like this:
`<source srcset="/img/my_img.png 135w">`

mertenhanisch:


fredrikekelund:
Thanks for filing a PR! Adding default support for srcset seems like a fair idea to me, but I think that, to align with the simplecrawler philosophy, we need to slim down this diff. Could we express this logic in a short function that's added to the discoverRegex's? Maybe by extracting the content of the srcset attribute, split by ",", clean up media queries, return?

fredrikekelund:
Also, would be great if you could add a small test :)

mertenhanisch:
That sounds good. I will need help with the regex, though.

PR #246, opened 2016-03-11
Added support for srcset attributes in discoverRegex

### LINKED ISSUE #237: Support srcset

fredrikekelund:
Fixes #237 and supersedes #239. Sorry about the coup @mertenhanisch, but I thought the easiest solution would be for me to implement the whole thing


mertenhanisch:
@fredrikekelund Hey, I won't complain. Thank you, sir!

Issue #88, opened 2014-08-01
Queue and QueueItem interfaces documentation


lzrski:
It would be nice to have `queue` and `queue item` interfaces documented for alternative implementations, e.g. #83

cgiffard:
Closing as partially resolved by #252 and the recent JSDoc PR.

PR #252, opened 2016-03-18
Big README update

### LINKED ISSUE #248: Troubleshooting Guide/FAQ
A lot of the same mistakes and questions keep cropping up repeatedly, and I think we should add a troubleshooting guide.

The big ones in my mind are:

* People wondering why the crawl is throwing complete without fetching any resources (usually it's that the first hit is a redirect to a domain not covered by the whitelist.)
* People wondering why the crawler picks up crappy URLs
* People confused about how to queue up their own resources (maybe this is really the docs needing clarification on this point)
* People who don't really get events or async (want to check a DB before queueing a result, but try and return the result synchronously, etc.)

fredrikekelund:
Cleaned things up a bit, made some parts more compact and tried to expand the information in others. Also made some corrections (eg. about the discoverResources method) and added an FAQ section as well as an index.

Fixes #248.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Simple web-crawler for Node.js
+# Simple web crawler for node.js
 
 [![NPM version](https://img.shields.io/npm/v/simplecrawler.svg)](https://www.npmjs.com/package/simplecrawler)
 [![Linux Build Status](https://img.shields.io/travis/cgiffard/node-simplecrawler/master.svg)](https://travis-ci.org/cgiffard/node-simplecrawler)
@@ -6,21 +6,10 @@
 [![Dependency Status](https://img.shields.io/david/cgiffard/node-simplecrawler.svg)](https://david-dm.org/cgiffard/node-simplecrawler)
 [![devDependency Status](https://img.shields.io/david/dev/cgiffard/node-simplecrawler.svg)](https://david-dm.org/cgiffard/node-simplecrawler#info=devDependencies)
 
-Simplecrawler is designed to provide the most basic possible API for crawling
-websites, while being as flexible and robust as possible. I wrote simplecrawler
-to archive, analyse, and search some very large websites. It has happily chewed
-through 50,000 pages and written tens of gigabytes to disk without issue.
-
-#### Example (simple mode)
-
-```js
-var Crawler = require("simplecrawler");
-
-Crawler.crawl("http://example.com/")
-    .on("fetchcomplete", function(queueItem) {
-        console.log("Completed fetching resource:", queueItem.url);
-    });
-```
+simplecrawler is designed to provide a basic, flexible and robust API for
+crawling websites. I wrote simplecrawler to archive, analyse, and search some
+very large websites. It has happily chewed through hundreds of thousands of
+pages and written tens of gigabytes to disk without issue.
 
 ### What does simplecrawler do?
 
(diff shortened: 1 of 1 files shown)


fredrikekelund:
This update also kinda concerns #88, since it updates the queue item documentation (which was previously incomplete). I'll leave the question of a more technical queue interface spec for another time, though.

README.md
```diff
@@ -6,21 +6,10 @@
 [![Dependency Status](https://img.shields.io/david/cgiffard/node-simplecrawler.svg)](https://david-dm.org/cgiffard/node-simplecrawler)
 [![devDependency Status](https://img.shields.io/david/dev/cgiffard/node-simplecrawler.svg)](https://david-dm.org/cgiffard/node-simplecrawler#info=devDependencies)
 
-Simplecrawler is designed to provide the most basic possible API for crawling
-websites, while being as flexible and robust as possible. I wrote simplecrawler
-to archive, analyse, and search some very large websites. It has happily chewed
-through 50,000 pages and written tens of gigabytes to disk without issue.
-
-#### Example (simple mode)
```
cgiffard: I reckon throwing some kind of basic code example at the very top might be useful, just to help people evaluate whether this is the right package for them, at a glance.

README.md
```diff
@@ -246,168 +246,188 @@ crawler.on("fetchcomplete", function(queueItem, data, res) {
...
-*    `crawler.stripQuerystring` -
-    Specify to strip querystring parameters from URLs. Defaults to false.
-*    `crawler.discoverResources` -
-    Use simplecrawler's internal resource discovery function. You can replace it
-    with your own function, which must accept a buffer and a queueItem, and add
-    the discovered resources to the crawler queue:
+* `crawler.stripQuerystring=false` -
+    Specify to strip querystring parameters from URL's.
+* `crawler.discoverResources` -
+    simplecrawler's built-in discovery function is purposefully naïve. The
```
cgiffard: I wonder if we might create a dedicated in the section to talk about link discovery, and link to it from here?

README.md
```diff
@@ -552,120 +569,153 @@ times a second. You can test the following properties:
...
-status at any one time, and/or retrieve them. That's easy with
+You'll probably often need to determine how many queue items have a given status
+and/or retrieve them. That's easily done with the methods
 `crawler.queue.countWithStatus` and `crawler.queue.getWithStatus`.
 
-`crawler.queue.countWithStatus` returns the number of queued items with a given
+`crawler.queue.countWithStatus` provides the number of queued items with a given
 status, while `crawler.queue.getWithStatus` returns an array of the queue items
-themselves.
+themselves. Again, by default, these methods both return and accept callbacks.
```
cgiffard: `return and accept callbacks`, should probably be just `take callbacks`. I kinda feel like erasing the existence of the synchronous methods since they're totally legacy and people shouldn't be using them. :)

README.md
```diff
@@ -552,120 +569,153 @@ times a second. You can test the following properties:
...
-    info was wrongly specified in requests]
-    (https://github.com/cgiffard/node-simplecrawler/pull/40)
-    and for fixing the missing request timeout handling!
-*   [Graham Hutchinson](https://github.com/ghhutch) for adding
-    querystring-stripping option
-*   [Jellyfrog](https://github.com/jellyfrog) for assisting in diagnosing some
-    nasty EventEmitter issues.
-*   [Brian Moeskau](https://github.com/bmoeskau) for helping to fix the confusing
-    'async' events API, and providing invaluable feedback.
+* [Nick Crohn](https://github.com/ncrohn) for the HTTP Basic auth support, and
```
cgiffard: I feel like we should generate a list of contributors from git — since there's heaps of people who are presently unacknowledged (including yourself, @fredrikekelund!) and I feel guilty about it. :)

Or maybe it'll be unwieldy in which case we should just replace it with a generic thanks message and link off to the contributor graph.

README.md
```diff
@@ -552,120 +569,153 @@ times a second. You can test the following properties:
...
 * `removecookie` (cookie array)
 Fired when one or more cookies are removed from the jar.
 
+## FAQ
+
+There are a couple of questions that pop up more often than others in the issue
+tracker. If you're having trouble with simplecrawler, please have a look at the
+list below before submitting an issue.
+
+- **Q: Why does simplecrawler discover so many crappy URL's?**
```
cgiffard: These look good! I hope you don't mind, but I might actually play around with the language here as it's all technically correct but feels a /tad/ unnatural to me (I'm a pedant, haha — please don't read too much into it!)

cgiffard:
Looks rad. I'm gunna merge this and then apply a few small changes on top myself. :)

README.md
fredrikekelund: That sounds like a good idea! Seeing how often the topic comes up, I think we want to make it easy to find info about

README.md
fredrikekelund: The reason I removed it was because of its brevity compared to how extensive simplecrawler's set of options are, and the slightly more extensive code examples are very near if one scrolls down just a bit. Still, maybe you're right - it might be a good way to get an instant picture of the module, even if it's very brief

README.md
fredrikekelund: Sounds like a good idea! I'll open a ticket for it, feels like a good change to include for a 1.0 release since it's API breaking

README.md
fredrikekelund: No worries :) I think maybe this section could be turned into some contributor guidelines, encouraging people to submit issues and PR's, and thanking all of those who've done it (not by name, but just like "thanks to all of those who've contributed"). Could be a link to the contributor graph as well!

README.md
fredrikekelund: Not at all, harmonizing the language throughout the docs should make it more readable!

PR #255, opened 2016-03-23
Redirects should not change the depth on the initial request

thejoshcrawford:
If allowInitialDomainChange is true crawls with maxDepth=1 will fail if depth is not reset.


cgiffard:
Looks cool — would you mind adding a test for your desired behaviour? :)

Thanks!

thejoshcrawford:
Turns out that 2xx responses weren't setting _isFirstRequest to false and redirects further down the chain could cause a domain redirect if it was allowInitialDomainChange was true.

Also the depth problem was only happening if two redirects happened in a row. Sort of an edge case, but I did notice it in this type of scenario http -> https -> blogspot.com

fredrikekelund:
Looks good to me! Thanks for adding the tests :+1:

PR #256, opened 2016-03-26
Updated queue methods to only work asynchronously

### LINKED ISSUE #253: Make queue methods only async
Even though the operations in the default queue methods are actually synchronous, the API is meant to be compatible with different queue backing stores that can perform async operations in those methods.

fredrikekelund:
Even though I didn't have to, some of the methods now use (faux)-asynchronous iteration to be compatible with other queue backing stores. If someone implemented another store, they'd probably be better off re-writing all methods themselves, but this might provide a better boilerplate.

I've also added unit tests for each queue method and updated the docs. Fixes #253.


lib/queue.js
```diff
@@ -26,16 +28,9 @@ module.exports = FetchQueue;
...
-    // For legacy reasons
-    if (depth instanceof Function) {
-        callback = depth;
-        depth = 1;
-    }
-
     depth = depth || 1;
-    callback = callback && callback instanceof Function ? callback : function() {};
-    var self = this;
+    callback = callback instanceof Function ? callback : noop;
```
fredrikekelund: I only made this change due to the previous syntax screwing up syntax highlighting for me :grin:

lib/queue.js
```diff
@@ -26,16 +28,9 @@ module.exports = FetchQueue;
...
 FetchQueue.prototype = [];
 FetchQueue.prototype.add = function(protocol, domain, port, path, depth, callback) {
-
-    // For legacy reasons
-    if (depth instanceof Function) {
-        callback = depth;
-        depth = 1;
-    }
-
     depth = depth || 1;
```
cgiffard: If the block above is removed, legacy support will be broken (I think it should be OK, but just wanted a second opinion.)

lib/queue.js
```diff
@@ -46,255 +41,242 @@ FetchQueue.prototype.add = function(protocol, domain, port, path, depth, callbac
...
     }
+};
 
-    this.scanIndex[url] = true;
-    callback(null, 0);
-    return 0;
-
+// Get item from queue
+FetchQueue.prototype.get = function(index, callback) {
+    callback = callback instanceof Function ? callback : noop;
```
cgiffard: Do we want to remove the error checking around the index param? Curious about the rationale there. (My preference would be to callback with error — the old code isn't wonderful there either.)

lib/queue.js
fredrikekelund: Oh yeah right, forgot to leave a comment about this! I figured that since the next release target probably is 1.0.0, it might be a good time to clean up any legacy code that's not documented etc. So long as we make sure to make a note of it when releasing

lib/queue.js
```diff
@@ -46,255 +41,242 @@ FetchQueue.prototype.add = function(protocol, domain, port, path, depth, callbac
...
-            self.oldestUnfetchedIndex = itemIndex;
-            item = self[itemIndex];
-            callback(null, item);
-            return item;
+    callback = callback instanceof Function ? callback : noop;
+    var queue = this;
+
+    queue.getLength(function(error, length) {
+        function getItem (itemIndex) {
+            queue.get(itemIndex, function(error, item) {
```
cgiffard: I'm wondering if it's possible we might blow the call stack if we're using recursion on queues with >65535 items?

Given this update is more about maintaining signature compatibility with async stuff as opposed to making it async itself, and the implementation is more or less opaque, I wonder whether a loop would be better? (Feel free to come back with a better reason to maintain the current strategy — just wanna talk this one through.)

lib/queue.js
```diff
@@ -46,255 +41,242 @@ FetchQueue.prototype.add = function(protocol, domain, port, path, depth, callbac
...
+                if (error) {
+                    callback(error);
+                } else if (item.status === "queued") {
+                    queue.oldestUnfetchedIndex = itemIndex;
+                    callback(null, item);
+                } else if (itemIndex + 1 < length) {
+                    getItem(itemIndex + 1);
+                } else {
+                    callback(new Error("No unfetched items remain."));
```
cgiffard: The fact that we call back with error here is a bit odd (and I'm wondering why I did it like that in the first place.)

Could we call back with (null, null)? Would that be understandable? Or is it truly an error condition?

lib/queue.js
```diff
@@ -46,255 +41,242 @@ FetchQueue.prototype.add = function(protocol, domain, port, path, depth, callbac
...
+    if (allowedStatistics.indexOf(statisticName) === -1) {
+        return callback(new Error("Invalid statistic"));
     }
 
-    self.forEach(function(item) {
-        if (item.fetched && item.stateData[statisticName] !== null && !isNaN(item.stateData[statisticName])) {
-            NumberSum += item.stateData[statisticName];
-            NumberCount++;
+    queue.forEach(function(item) {
+        if (item.fetched && Number.isFinite(item.stateData[statisticName])) {
```
cgiffard: Bear in mind this is an ES2015 function. I checked node 0.10 and it seems OK, but I'm wondering whether we should set a base support level for the crawler (thoughts?)

lib/queue.js
```diff
@@ -46,255 +41,242 @@ FetchQueue.prototype.add = function(protocol, domain, port, path, depth, callbac
...
-    if (allowedStatistics.join().indexOf(statisticName) === -1) {
-        // Not a recognised statistic!
-        return callback(new Error("Invalid statistic."));
+    callback = callback instanceof Function ? callback : noop;
+
+    var sum = 0,
+        count = 0,
+        queue = this;
+
+    if (allowedStatistics.indexOf(statisticName) === -1) {
```
cgiffard: Nice catch. That was dumb of me. :P

test/queue.js
```diff
@@ -0,0 +1,137 @@
+/* eslint-env mocha */
```
cgiffard: Epic.

cgiffard:
Looks really good! I left some comments for your perusal. :)

lib/queue.js
cgiffard: Yeah, that's cool, then. :)

lib/queue.js
fredrikekelund: You're probably right about this one, I think we ought to revert it. While it is unlikely that it would actually iterate over that many items in this method no matter how lang the queue is, it shouldn't pose a problem if it had to.

I think this change mostly came out of me trying to go all in with the async methodology, but since the store isn't async, you're absolutely right - we don't need to adhere to those constraints here, especially if it risks blowing the call stack (which I didn't know had that limit btw, good to know!)

lib/queue.js
fredrikekelund: I was unsure of node 0.10 support at first as well, thought I'd rely on the CI to see how it did, and it passed. That's what the module requires as the lowest supported version in the package.json as well, so I figure it's no problem if we make use of it. Whatcha think?

lib/queue.js
fredrikekelund: Hmm you might be right that it's more convenient for the developer to get back an error when the index parameter is out of range. I figured that maybe the developer should keep track of the range themselves, but if we verify the range, then the developer can be sure that the second callback parameter is always a queue item so long as there's no error. I'll revert this!

lib/queue.js
```diff
@@ -46,255 +41,242 @@ FetchQueue.prototype.add = function(protocol, domain, port, path, depth, callbac
...
-    return 0;
-
+// Get item from queue
+FetchQueue.prototype.get = function(index, callback) {
+    callback = callback instanceof Function ? callback : noop;
+    callback(null, this[index]);
 };
 
 // Get last item in queue...
 FetchQueue.prototype.last = function(callback) {
```
fredrikekelund: Is this method used anywhere? It might provide some convenience, but I'm not sure in what scenario. You think it's something we should consider dropping, or is it worth keeping @cgiffard?

lib/queue.js
cgiffard: I think when reverting we should call back with an error — it should be clear you're getting an index which is out of range.

lib/queue.js
cgiffard: Originally it was intended to be part of the public API, in case anybody wanted to use it. If nobody is, we can get rid of it (but it's so hard to know. :-/)

cgiffard: The hard part is most of the projects using simplecrawler are private, I think. There's a few link checkers and stuff on npm, but most of the crazy stuff (that all the people riled up about robots.txt support are seeing) would be run by individuals for their own means.

So it's really hard to know what people are using. :-/

lib/queue.js
cgiffard: I also think it's pretty unlikely, but with a loop we can avoid the possibility altogether. :)

lib/queue.js
cgiffard: I'm happy with that. Let's make 0.10 our lowest supported version from here-on-in. :)

lib/queue.js
fredrikekelund: Difficult one. It feels a bit odd to call back with nothing, neither an error nor a queue item, but I agree that this isn't really an error condition either. I'll change it to call back with (null, null) and let's make sure to make a note of this in the API docs once we've settled on a way to write those

lib/queue.js
fredrikekelund: I've changed this now so that we call back with a RangeError. Also added a unit test for it

lib/queue.js
```diff
@@ -46,255 +41,246 @@ FetchQueue.prototype.add = function(protocol, domain, port, path, depth, callbac
...
     });
-    average = NumberSum / NumberCount;
-    callback(null, average);
-    return average;
+
+    callback(null, sum / count);
 };
 
 // Gets the number of requests which have been completed.
 FetchQueue.prototype.complete = function(callback) {
```
fredrikekelund: While we're at it, I'd like to consider renaming this method `getCompletedCount`. It's more in line with `countWithStatus`, and I think the current name is quite confusing.

We could keep the old method as an alias and spit out a deprecation warning to the console whenever it's used

lib/queue.js
```diff
@@ -26,16 +28,9 @@ module.exports = FetchQueue;
 
 FetchQueue.prototype = [];
 FetchQueue.prototype.add = function(protocol, domain, port, path, depth, callback) {
```
fredrikekelund: How do you feel about changing the signature of this function to accept just two arguments: `queueItem` and `callback`, @cgiffard? It would make it more intuitive to call and since we construct a queue item from the arguments anyway, it would let us shorten the method without having to add much code at all in the places where it's called

fredrikekelund:
I'd like to consider adding two new methods to the queue: `remove` and `requeue`. `remove` would actually remove a queueItem from the queue, whereas `requeue` would add a URL and bypass any checks whether a URL has already been fetched or not before adding it.

cgiffard:
I totally get the utility of `requeue` — a resource might've failed previously, or you might want to download it again to check for changes. But not sure about `remove` — would that just be for truncating the queue to save memory? (Not saying we shouldn't add it, but it helps to understand the intent behind a feature in order to design the API properly.)

cgiffard:
So I got some comments via email:

> Hi Christopher,
> 
> The only thing I've spotted so far is that it appears the change of state is a race condition: you emit say 'fetchstart' but you have no way to wait for the, possibly asynchronous, operation to complete (that I've seen thus far). That may run into trouble just with a single crawler, but shared use of a single queue is essentially impossible without additional handshaking in the background - really I'd just like to rely on transactions (I'm using sequelize) to see if I 'own' the state change and communicate back a proceed or abort to next flag of some kind.
> 
> Perhaps I'm looking too hard at the mongo imp. which basically acts on those emits - I think at least some of them can be worked into the fetchQueue interface (but only in a hacky, knowledge of how called way). An explicit changeState interface in fetchQueue would be much preferable.
> 
> Regards
> 
> Pete

Then I said:

> Hi Pete —
> 
> Yeah, @frederikekelund suggested an additional ‘upsert’ method on the queue for the async-safe modification of queue items might be the way to go. I’ll file an issue for it and we can go from there!
> 
> Thanks for looking into this. :)
> 
> Cheers,
> 
> Christopher

fredrikekelund:
Cool to see people working on queue implementations! Can we convince him to share his code at some point? :)

fredrikekelund:
I gave some thought to the `remove ` method as well, and I initially imagined it mostly as a natural complement to any form of collection that also has an `add ` method. But I agree that the need to remove stuff is rare, instead you want to prevent stuff from being added to the queue, not remove stuff once it's already there.

I think that the best use case I could think of is one where you've already fetched a page, but you know you'll encounter it again at some point (perhaps after some cookie has been set on a different page), and you don't want it to be prevented from being added to the queue at that point. But that situation might be more elegantly solved with an `unset` method that doesn't remove any existing queue items, but rather just changes the state of the URL in the scanIndex

fredrikekelund:
On the discussion of removing the `last` method as well: I think that we can afford to cut out methods that aren't used in simplecrawler, especially when their implementation can be redone manually without much effort by anyone who really needs the functionality. I'm not sure of a typical use case of that method either

lib/queue.js
cgiffard: I wonder if we need a handy public method for constructing a queueItem from a URL, in that case? So you'd do something like:

```js
crawler.queue.add(crawler.queue.createQueueItem("http://foo.com:8080/bar?baz"), function() {
    // stuff here
});
```

cgiffard:
I think this looks pretty good. I might actually write a redis queue implementation to validate the queue interface's asynchronicity here. :)

lib/queue.js
fredrikekelund: Since the `crawler.processURL` method basically does the same thing as the proposed `crawler.queue.createQueueItem`, this raises a good point about compatibility between the recurring but undocumented `parsedURL` type object and `queueItem`s.

I think what we ought to change here is to have `crawler.processURL` construct a full `queueItem` (which would mean adding the properties `url`, `depth`, `fetched`, `status` and `stateData` to the returned object) and then have the `add` method basically just push it to the queue. I was thinking about having the `add` method set a `queueItem.id` property as well, but considering the fact that URL's are unique, the `queueItem.url` property should be able to fill that function for queue implementations that need to look up the queue item in eg. a database.

fredrikekelund:
Alright! Haven't had time to look into this for some time, but now I'm back :)

I've removed the `last` method and added a `force` argument to the `queue.add` method. Moreover I made `crawler.processURL` construct a full `queueItem` rather than an incomplete one (ie. a `parsedURL`). This should hopefully add some clarity from the developers perspective and it gives the `processURL` method a purpose to provide standardized queueItems to those who want to write custom queue implementations as well.

I should have a second look at the diff during the coming days, but conceptually I feel pretty good about the changes now

cgiffard:
I think we could add a `requeue` to the `Crawler` prototype which would map to `queue.add` with `force` applied.

That'd solve issue #286. :)

cgiffard:
That's probably a separate PR though, but it's blocked on this one. I'll have a look over this diff, and if you don't feel bad about it, let's merge this one for 1.0!

fredrikekelund:
A `requeue` method on `Crawler.prototype` sounds good to me. I've made some more changes to harmonize the names of the queue methods a bit. Wondering whether we should change `countWithStatus` to `getStatusCount` while we're at it - just to harmonize things a bit?

cgiffard:
Hrm. What's the pattern you'd like to follow? At first glance, I find `countWithStatus` a little clearer than `getStatusCount`. Maybe `getStatusCount` or `getCountForStatus` are clearer.

(Because they look more like a sentence when you've got them in a function call — i.e. `getCountForStatus("queued", ...`, etc.)

fredrikekelund:
Maybe it's overkill to prepend all method names with "get", you might be right about the current names being the most intuitive.

Just to open up for discussion: would it make sense to generalize the queue methods `countWithStatus` and `getWithStatus` into something like `countItems` and `filterItems`, that would take an object comparator. So it would look something like:
```js
crawler.queue.countItems({status: "fetched"});
```

With good documentation on what the queue items look like, these methods might be even more useful while still being pretty elegant to call. What do you think?

fredrikekelund:
@cgiffard I went ahead with the `queue.filterItems` and `queue.countItems` change. So long as the documentation for what the queue items look like is good enough, I think this concept serves us better as an API, seeing as it's more flexible.

I'd like to go ahead and merge this PR this weekend unless you have any objections?

cgiffard:
Having a look now! :) Thanks @fredrikekelund!

lib/crawler.js
```diff
@@ -1378,21 +1380,21 @@ Crawler.prototype.handleResponse = function(queueItem, response, timeCommenced)
...
         }
 
         if (crawler.allowInitialDomainChange && crawler._isFirstRequest) {
-            crawler.host = parsedURL.host;
+            crawler.host = newQueueItem.host;
         }
 
         // Clean URL, add to queue...
-        crawler.queueURL(parsedURL, queueItem);
+        crawler.queueURL(newQueueItem, queueItem);
```
cgiffard: Maybe we can rename `queueItem` here to be 'referringQueueItem` or something, to make it clearer.

lib/queue.js
```diff
@@ -16,6 +16,29 @@ var allowedStatistics = [
...
+ */
+function compare (a, b) {
+    var matches;
+
+    for (var key in a) {
+        if (a.hasOwnProperty(key)) {
+            if (typeof a[key] === "object" && typeof b[key] === "object") {
+                matches = compare(a[key], b[key]);
+            } else {
+                matches = matches === false ? matches : a[key] === b[key];
```
cgiffard: I think we could bail out early here. If any of the matches come back false, we should immediately return false rather than continuing to loop. :)

lib/queue.js
```diff
@@ -25,276 +48,197 @@ var FetchQueue = function() {
...
-FetchQueue.prototype.add = function(protocol, domain, port, path, depth, callback) {
+FetchQueue.prototype.add = function(queueItem, force, callback) {
+    callback = callback instanceof Function ? callback : noop;
+    var queue = this;
 
-    // For legacy reasons
-    if (depth instanceof Function) {
-        callback = depth;
-        depth = 1;
+    if (arguments.length === 2) {
```
cgiffard: I wonder if we should leave this function with a pseudo-variadic calling style, or just make people send `force`. I'm leaning towards the latter, since we already provide a much friendlier queueing method on the crawler itself. But happy to hear arguments to the contrary!

lib/queue.js
```diff
@@ -25,276 +48,197 @@ var FetchQueue = function() {
...
-
-                self.push(queueItem);
-                callback(null, queueItem);
+    queue.exists(queueItem.url, function(err, exists) {
+        if (err) {
+            callback(err);
+        } else if (!exists) {
+            addToQueue();
+        } else if (force) {
+            if (queue.indexOf(queueItem) > -1) {
```
cgiffard: Good catch — at least with the current queue implementation, not having this condition would mean that statuses could be updated for more than one queue item simultaneously, completely erroneously.

lib/queue.js
```diff
@@ -25,276 +48,197 @@ var FetchQueue = function() {
...
-                var error = new Error("Resource already exists in queue!");
-                error.code = "DUP";
-
-                callback(error);
+                addToQueue();
             }
-        });
+        } else {
+            var error = new Error("Resource already exists in queue!");
+            error.code = "DUP";
```
cgiffard: Wondering whether we might change `error.code` to read `DUPLICATE`, just for clarity. It's not like we're getting taxed on the extra letters. :)

(It was my fault, sorry!)

cgiffard:
Looks great! I really appreciate the work you've put in here — should massively improve the experience of building DB backed queues. I might write a bunch of content for the README explaining how one would go about writing a queue.

Left a couple of comments with thoughts — feel free to address these at your leisure. I don't think they should block merging! This is really great — it'll be fantastic to get this in. :)

lib/queue.js
fredrikekelund: Yeah, you're probably right! I think the thought might've struck me, but I skipped it since I figured the comparator object should always be fairly small. But it'd probably smarter to just optimize for performance, especially for a project like this that serves as the base for other ones.

lib/queue.js
fredrikekelund: True, we shouldn't encourage people to call this method directly, so let's always require the same signature

lib/queue.js
fredrikekelund: Good call, I think we can afford the extra letters 😉

fredrikekelund:
Cool! Yeah, feels like a more formal API with these changes. Would be great to document the process more. I'll look into the last points this week!

Do you feel that we still need to cut a 0.8 release before 1.0? I'm thinking that we ought to just push ahead. If you don't have time to finish up the CLI, then that can be added in an upcoming release, it shouldn't be API breaking, so we don't have to change the major version number when that lands.

cgiffard:
Fuck it, let's just cut a 1.0. :)

I wanna get the CLI out, but if you're itching to release 1.0 we can always push it as 1.1, since it won't break any of the existing APIs. I won't be able to get to it before the weekend unfortunately.

fredrikekelund:
A bit of an itch :) The robots.txt stuff has been sitting around for so long, and I really want to break that `fetchWhitelistedMimeTypesBelowMax` API haha.

I'll try and get to the remaining stuff during the week, and we can aim for a release this Sunday?

cgiffard:
Sounds good!

cgiffard:
Awwwww yeah! Thanks @fredrikekelund! :)

PR #261, opened 2016-04-02
Decompress compressed responses before emitting fetchcomplete

### LINKED ISSUE #260: Can't convert response buffer to a string when response is gzipped
I'm trying to convert the response buffer to a string but it seems that when response is gzipped, simplecrawler can't decode it.

Here is my playground: 

```
var Crawler = require("simplecrawler");
var crawler = new Crawler("elliptips.info", "/", 80, 100);

crawler.protocol = 'https';
crawler.decodeResponses = true;
crawler.customHeaders = {
    // don't use gzip for testing
    'accept-encoding': null
}
crawler
    .on("crawlstart", function() {
        console.log("Crawl starting");
    })
    .on("gziperror", function(queueItem, error, resourceData) {
        console.log("gzip error: ", error);
    })
    .on("fetchcomplete", function(queueItem, responseBuffer, response) {
        console.log(response.headers);
        console.log(responseBuffer.toString());
        process.exit()
    })
    .on("complete", function() {
        console.log('crawl complete');
    });

crawler.start();
```

When I remove 'accept-encoding' to enable gzip, I get this :

capture d ecran 2016-04-01 a 15 25 42

I've tried with multiples websites and always get the same result.

I'm currently on master branch :

package.json
    ....
    "simplecrawler": "cgiffard/node-simplecrawler",
    ....

Edit: I get the same result with or without decodeResponse set to true.

fredrikekelund:
By default, responses that are gzipped or deflated are now
decompressed before they're emitted in the fetchcomplete event.
Previously they have only been decompressed before resource
discovery (and they are still always), but now the
`crawler.decompressResponses` setting (true by default) determines
whether the fetchcomplete response should be decompressed as well.

In theory this could boost performance, since we're doing the same
amount of computation as previously, but with less network
traffic.

Fixes #260

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,8 +279,13 @@ change to adapt it to your specific needs.
     The maximum time in milliseconds the crawler will wait for async listeners.
 * `crawler.userAgent="Node/simplecrawler <version> (https://github.com/cgiffard/node-simplecrawler)"` -
     The user agent the crawler will report.
+* `crawler.decompressResponses=true` -
+    Response bodies that are compressed will be automatically decompressed
+    before they're emitted in the `fetchcomplete` event. Even if this is falsy,
+    compressed responses will be decompressed before they're passed to the
+    `discoverResources` method.
 * `crawler.decodeResponses=false` -
-    The response bodies will be intelligently character converted to standard
+    Response bodies will be intelligently character converted to standard
     JavaScript strings using the
     [iconv-lite](https://www.npmjs.com/package/iconv-lite) module. The character
     encoding is interpreted from the Content-Type header firstly, and secondly
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -89,6 +89,10 @@ var Crawler = function(host, initialPath, initialPort, interval) {
     // Should we update crawler.host if the first response is a redirect to another domain.
     crawler.allowInitialDomainChange = false;
 
+    // Set Accept-Encoding header and automatically decompress HTTP responses
+    // based on Content-Encoding header
+    crawler.decompressResponses = true;
+
     // Decode HTTP responses based on their Content-Type header or any
     // inline charset definition
     crawler.decodeResponses = false;
@@ -381,11 +385,14 @@ Crawler.prototype.getRequestOptions = function (queueItem) {
         agent: agent,
         headers: {
             "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
-            "Accept-Encoding": "gzip, deflate",
             "User-Agent": crawler.userAgent,
             "Host": queueItem.host + (isStandardHTTPPort || isStandardHTTPSPort ? ":" + queueItem.port : "")
         }
     };
+
+    if (crawler.decompressResponses) {
+        requestOptions.headers["Accept-Encoding"] = "gzip, deflate";
+    }
 
     if (queueItem.referrer) {
         requestOptions.headers.Referer = queueItem.referrer;
@@ -937,13 +944,11 @@ Crawler.prototype.domainValid = function(host) {
     one, since this function queues the resources in addition to
     discovering them.
 
-    resourceData    - Text document containing linked resource URLs.
-    queueItem       - Queue item from which the resource document was derived.
-    decompressed    - Content is already decompressed (default: false)
+    resourceData - Text document containing linked resource URLs.
+    queueItem    - Queue item from which the resource document was derived.
 
     Emits
 
-        gziperr
         discoverycomplete
 
     Examples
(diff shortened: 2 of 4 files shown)


fredrikekelund:
Have you had a chance to have a look at the diff here, @cgiffard? :)

fredrikekelund:
Don't sweat it too much :) I might work on some other issues tomorrow, but once we've settled on this one, we can cut a 0.7.0 release, would be cool if we could do that this weekend!

fredrikekelund:
Inclined to merge this today and release 0.7.0. Objections, @cgiffard?

README.md
```diff
@@ -279,8 +279,13 @@ change to adapt it to your specific needs.
     The maximum time in milliseconds the crawler will wait for async listeners.
 * `crawler.userAgent="Node/simplecrawler <version> (https://github.com/cgiffard/node-simplecrawler)"` -
     The user agent the crawler will report.
+* `crawler.decompressResponses=true` -
```
cgiffard: I'm actually starting to wonder whether this should even be configurable. Is there a scenario where we'd ever want to switch decompression off? (Ideally I'd like to make sure that the number of configuration options we have to maintain into the distant future is considerably reduced.)

lib/crawler.js
```diff
@@ -1271,17 +1273,23 @@ Crawler.prototype.handleResponse = function(queueItem, response, timeCommenced)
...
+            // because we always provide the discoverResources method with a
+            // decompressed buffer
+            if (/(gzip|deflate)/.test(stateData.headers["content-encoding"])) {
+                zlib.unzip(responseBuffer, function(error, decompressedBuffer) {
+                    if (error) {
+                        crawler.emit("gziperror", queueItem, error, responseBuffer);
+                        emitFetchComplete(responseBuffer);
+                    } else {
+                        var responseBody = crawler.decompressResponses ? decompressedBuffer : responseBuffer;
+                        emitFetchComplete(responseBody, decompressedBuffer);
```
cgiffard: I just noticed that in the scenario where the body is successfully decoded, we're passing the decompressed response body twice. Is this something we need to do? (If we don't make it configurable, we can avoid that weird logic altogether!) If we decide to leave it in, let's comment it.

lib/crawler.js
```diff
@@ -1241,11 +1234,20 @@ Crawler.prototype.handleResponse = function(queueItem, response, timeCommenced)
...
 
     // Ensure response length is reasonable...
-    responseLength =
-        responseLength > 0 ? responseLength : crawler.maxResourceSize;
-
+    responseLength = responseLength > 0 ? responseLength : crawler.maxResourceSize;
     queueItem.stateData.contentLength = responseLength;
 
+    function emitFetchComplete(responseBody, decompressedBuffer) {
+        responseBody = crawler.decodeResponses ? crawler.decodeBuffer(responseBody, stateData.contentType) : responseBody;
```
cgiffard: On the flipside, there's definitely a reason for us to allow switching off response decoding — after all, iconv-lite might get it wrong (and has for me quite frequently) — reducing the page to gibberish.

lib/crawler.js
```diff
@@ -381,12 +385,15 @@ Crawler.prototype.getRequestOptions = function (queueItem) {
         agent: agent,
         headers: {
             "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
-            "Accept-Encoding": "gzip, deflate",
```
cgiffard: And it looks like we were always sending this header, so the expected response is probably just to decompress by default.

cgiffard:
@fredrikekelund Sorry for the delay. Left a couple of comments. Let's address them and then I think we're good to go!

README.md
fredrikekelund: Good point, I think it'd be pretty sound to automatically decode responses. Could either change the default for 1.0.0 or just go ahead and remove the option. I'm not sure what happens to binary resources today, though. Might be worth looking into to make sure it's done logically.

lib/crawler.js
fredrikekelund: Yes, I added that in d5d53a1a2559f90cac96d8abb384b78fe64fb2fa because I thought we actually did decompress responses based on their headers, but turns out that was only done for the discoverResources method.

That change hasn't made it into a release yet, but I think that setting `decompressResponses=true` as a default is smart.

lib/crawler.js
fredrikekelund: Ah, really? I don't think I've had much trouble with it. Do you know of any particular scenarios where it's caused problems?

lib/crawler.js
fredrikekelund: Agreed that things get a bit convoluted here, but yeah, it's due to the flexibility in the config params. I agree with your original point that for 1.0.0 I think we should consider not making `decompressResponses` configurable, so long as we make sure we deal with binary resources correctly.

fredrikekelund:
Was there anything here that needed to be addressed in the code before merging, or mostly just things related to how we deal with decompression of responses in future versions?

robinparisi:
Hi,

Any news on the 0.7 release ?

lib/crawler.js
cgiffard: Last time I ran it it didn't have great support for SHIFT-JIS and EUC-JP, but that might be better now.

cgiffard:
@robinparisi It's released! :)

Issue #266, opened 2016-04-05
only fetch html content without any resources


hbakhtiyor:
i want only fetch html content without any resources, like js, css, images, pdf, ....
how to do?

fredrikekelund:
First, set `crawler.supportedMimeTypes=["text/html"]`, then set `crawler.downloadUnsupported=false`. You can also write a fetch conditions that prevents URL's that end in eg. ".jpg", ".css", ".js", ".pdf" or ".png" in order to not even try and fetch those URL's (for resources that are discarded based on mimetypes, the crawler still has to fetch headers to determine what mimetype it's about to receive)

hbakhtiyor:
thanks,

btw, `supportedMimeTypes` option accepts regex object, also need to accept string too, it will be more flexible.

cgiffard:
You match anything with a regex you can match with a string. :)

PR #268, opened 2016-04-06
Add accepting string type for supportedMimeTypes option

hbakhtiyor:


cgiffard:
It is my preference to keep the API simple rather than accepting matches in a million different formats, but happy to discuss it. Thoughts, @fredrikekelund?

fredrikekelund:
I think there's value in using RegExp over strings (even though I got it mixed up when commenting on #266) because it makes it easy to match all image types, or all text types. And even if you're not super familiar with RegExp, you can basically just swap the quotes around a string for forward slashes and get the same result

cgiffard:
Yeah, agreed. I'd prefer not to complicate the API for this.

I do appreciate the effort involved in submitting the PR though, @hbakhtiyor!

Issue #270, opened 2016-04-09
Crawler hangs right before the end of the article


thecameronjones:
I am using this software to crawl Wikipedia as a test. The repository with my code can be seen [here](https://github.com/thecameronjones/crawler). I am using a crawler depth of 2 so it will only read the articles linked to the main one initially specified. Right before it finishes reading the end of the initial article, it will hang and the `complete` event won't fire. Also, there will still be a couple links left at the end of the article that it didn't crawl. I have tried many times to debug my code and fix this issue but I have had no success. Does anyone know if my code is the issue or if it's a legitimate bug?

fredrikekelund:
It looks like you forgot to call toString on the buffer [here](https://github.com/thecameronjones/crawler/blob/master/app.js#L31). Also keep in mind that if you load the response body as a cheerio instance twice and modify one of them, that won't affect the other instance. You can load it once and attach the instance to the current queueItem to be able to pass it between event listeners and discoverResources. Just remember to unset it when you're done so it can be picked up for GC. Otherwise your app will be using a lot more memory than necessary

thecameronjones:
Thanks for the reply. Calling `toString` does not seem to solve the problem. It still hangs a couple links before it's supposed to stop. Also, as I am a novice using cheerio (the jquery-like format) and optimizing memory usage in code, could you be _more_ specific about what you mentioned? Although the primary issue is still the crawler hanging. @fredrikekelund

cgiffard:
Hey! Would you mind adding this code to your crawler and letting me know what the output is? Often, the crawler might be doing something else that you won't see if you're only listening to `fetchcomplete`. (For now I can't rule out that it isn't a bug, but the output of this would help there too.)

```js
    var originalEmit = crawler.emit;
    crawler.emit = function(evtName, queueItem) {
        crawler.queue.complete(function(err, completeCount) {
            if (err) {
                throw err;
            }

            crawler.queue.getLength(function(err, length) {
                if (err) {
                    throw err;
                }

                console.log("fetched %d of %d — %d open requests, %d open listeners".green,
                    completeCount,
                    length,
                    crawler._openRequests,
                    crawler._openListeners);
            });
        });

        console.log(evtName, queueItem ? queueItem.url ? queueItem.url : queueItem : null);
        originalEmit.apply(crawler, arguments);
    };
```

(@frederikekelund I should add this to the troubleshooting section in the readme.)

thecameronjones:
[...]

Issue #249, opened 2016-03-15
Add a concept of download conditions


fredrikekelund:
In addition to fetch conditions, a powerful addition would be a flexible system for determining whether or not to fetch the body of a certain resource. I think we could call this concept download conditions.

The idea is similar to fetch conditions, you register handlers that are executed once the headers for a particular resource have been fetched. You the developer are free to synchronously evaluate these headers however you like, and the return value of the handler will determine whether to continue fetching the resource.

We could potentially fire a `fetchprevented` event for resources that are avoided due to download conditions.

This also means we should be able to drop the `fetchWhitelistedMimeTypesBelowMaxDepth` concept, which always seemed very specific to certain use cases to me.

fredrikekelund:
Working on this, should have a PR ready later today!

cgiffard:
Closed by #273. :)

PR #272, opened 2016-04-17
Add 'fetchdrop' event

Autarc:
Currently the crawler doesn't provide information about `queueItems` which were dropped because of the `maxDepth` limitation. The new event notifies in case a `fetchcomplete` is expected but not emitted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,9 +181,12 @@ in parentheses.
 * `cookieerror` (queueItem, error, setCookieHeader) -
     Fired when an error was caught trying to add a cookie to the cookie jar.
 * `fetchcomplete` (queueItem, responseBody, responseObject) -
-    Fired when the resource is completely downloaded. The response body is
+    Fired when a resource is completely downloaded. The response body is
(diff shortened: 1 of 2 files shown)


fredrikekelund:
Thanks for your contribution! The problems around the current maxDepth logic were recently up for discussion in #270 (look towards the end of the discussion). The fact that we need to fetch resources before checking their depth is due to the fetchWhitelistedMimeTypesBelowMaxDepth setting, which is about to be deprecated and is due for removal soon after that. 

I agree that this event is a way of bringing some much needed flexibility into the maxDepth logic, but I'm hesitant to adding this at this point, and would rather add the generalized download conditions ASAP (see #249). If you're inclined to, feel free to submit a PR with such an implementation btw :)

cgiffard:
@fredrikekelund I'm wondering whether we should extend `fetch disallowed`, or make a new event for resources which fail the crawler validity checking, regardless of reason. Then if we've got an equivalent for 'download conditions', we call it something similar.

PR #273, opened 2016-04-17
Add configuration option for 'whitelistedMimeTypes'

Autarc:
In addition to #272 (which was btw. just meant to assist developers to debug the current system) I think its useful to provide an option to configure the `whitelistedMimeTypes` for checking which data should be kept after fetching them. Currently these are [hardcoded in the library](https://github.com/cgiffard/node-simplecrawler/blob/master/lib/crawler.js#L614) and can't be modified (e.g. for checking the `text/html` content-type). The modification is focusing on backwards compatibility to existing code and therefore completely optional for cases its needed.

The issue #249 @fredrikekelund mentioned last time will probably cover similar scenarios but won't allow a seamless integration without changing parts of the internals.

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,9 +381,15 @@ change to adapt it to your specific needs.
     be downloaded. Asset files are excluded from this distance condition if
     `crawler.fetchWhitelistedMimeTypesBelowMaxDepth` is `true`. Defaults to `0` —
     no max depth.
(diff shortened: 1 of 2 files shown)


cgiffard:
Non-breaking change, pretty small. Target 0.7.0 for this one, I think.

PR #274, opened 2016-04-17
Use 'Array.protoype.some' instead of 'Array.prototype.reduce' for contains

Autarc:
For a better understanding whats happening and an increased performance, [Array.prototype.some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) is used instead of [Array.prototype.reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) for lookups.

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -564,9 +564,9 @@ Crawler.prototype.protocolSupported = function(URL) {
         return false;
     }
 
-    return crawler.allowedProtocols.reduce(function(prev, protocolCheck) {
-        return prev || !!protocolCheck.exec(protocol);
-    }, false);
+    return crawler.allowedProtocols.some(function(protocolCheck) {
+        return protocolCheck.test(protocol);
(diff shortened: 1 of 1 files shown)


cgiffard:
I'm more or less cool to merge this as I see no specific regressions. That said, I don't ever think a performance cost was remotely appreciable here (given these lists are tiny, we're talking ns in a workload which is mostly network bound anyway) — this update is purely useful for readability.

@fredrikekelund thoughts on the naming of the iterator? I'd been calling them `cur` through the codebase, for `current` — @Autarc renames them to `entry` here, which superficially I feel might be a tad ambiguous. Perhaps we can chose a new name? I'm fine with whatever, but consistency is king.

cgiffard:
After much deliberation — just merging this. We can bikeshed about iterator names later. :)

Issue #257, opened 2016-03-29
Asynchronously save queue items


fredrikekelund:
Building on #253, I'd like to look into how we mutate queue items. Currently it's all done synchronously, and I'd like to save them asynchronously after they're mutated, perhaps via an upsert method on the queue that takes two arguments, `id` and a `queueItem`. With this, we'd truly enable different queue backing stores, and even make it really easy to run the crawler in a distributed fashion.

cgiffard:
Good point — and something I'd previously missed. If we can make this work without complicating the queue interface, we're in a good position.

fredrikekelund:
I think so too, it struck me how easy it would be to run simplecrawler in a cluster if we had this.

And I don't think this would necessarily complicate the queue interface, but I do think Promises could provide a real benefit here. What are your thoughts on bringing in a dependency like Bluebird?

cgiffard:
Yeah, the intention behind a DB-backed queue was actually so you could run simplecrawler in a cluster. My preference with dependencies is actually to keep things really lean, and let consumers of the API wrap their code in promises if they want it — keeping callbacks underneath.

But, I think it would be easier to make the decision if we first identified all the code which is mutating the queue, and how easy it would be to write to a given pattern.

fredrikekelund:
Alright, I can sympathise with that. I'll see if I have time this weekend to bang out an initial implementation!

fredrikekelund:
Building this comment https://github.com/cgiffard/node-simplecrawler/pull/256#discussion_r63279012, I think that `crawler.queue.upsert` ought to take 2 arguments, a URL that works as an id, and an object that describes the changes to be made to the queueItem that represents that URL. In the default queue implementation, this basically means we'll use `Object.assign(originalQueueItem, updatedQueueItem)`

fredrikekelund:
I'll have a look at this tonight! This and #258 should be the remaining two things that need to land before 1.0.0 (besides writing some good release notes along with an upgrade path)

jtlapp:
I'm looking for a crawler that I can back with a robust queue on PostgreSQL. Looking over the queue interface just now, I'm wondering if it might require revision to support the necessary atomicity of queue transactions.

In particular, it seems to me that oldestUnfetchedItem() is the only query for which the crawler could trust results. Anything else is asking the queue for its current state, which might change before the state is subsequently acted on.

[...]

PR #283, opened 2016-05-23
Store the referrer in the queue

evenisse:


evenisse:
It is useful to store the referrer in the queue for specific queue implementations like DB queues and cluster environments.

fredrikekelund:
Hi, thanks for filing a PR! Exciting to see some interest in alternative queue implementations! It would indeed be useful to store the referrer URL in the queue, and it would need to be passed to the `add` method for it to work. But the add method's signature is changing as soon as #256 is merged (hopefully tomorrow), and so it feels a bit silly to merge this only to overwrite it almost immediately.

But please join the discussion in #253 and #257 for more on this topic!

Issue #208, opened 2016-01-26
Have crawler follor redirect Http 301


perrohunter:
Hi,
The crawler is hitting urls which were permanently moved and return status 301 on the fetcherror event.

Is there a way to have the crawler follow the redirect to the new location?

cgiffard:
Hi — thanks for filing an issue.

[The code in question should respect any 3XX response which sets a location header](https://github.com/cgiffard/node-simplecrawler/blob/master/lib/crawler.js#L1184-L1185)

Can you post a test case or some code which shows the problem? Could the server not be setting a `Location` header? Perhaps the redirect is to a hostname which is not whitelisted by your initial configuration?

tibetty:
I tried but the hang-up issue is still there... Really appreciate your quick response though!

fredrikekelund:
@tibetty do you mean the issue from #214? Would be nice if we could keep that discussion in that thread, you can still comment there, even if the ticket is closed. I'll go ahead and reopen it in any case

fredrikekelund:
If you're still having trouble with this, feel free to open a new issue, @perrohunter, but it's not unlikely that the problem doesn't lay with simplecrawler not following redirects (could maybe be due to some header not being sent or something similar - see #228 for a recent fix related to default header)

Issue #145, opened 2015-06-03
Crawler stops after redirect


teawithfruit:
I'm using simplecrawler for a project.
One of the websites to crawl have a redirect from the entry domain. Similar like when you type http://www.microsoft.com and you will get to http://www.microsoft.com/en-us/default.aspx.
In my opinion the redirected url should be added for crawling to the queue but the crawler stops, if the the entry domain is a redirect.

Here you can see my crawler code.
```javascript
var theCrawler = simplecrawler.crawl('http://www.microsoft.com')
.on("fetchcomplete", function(queueItem, responseBuffer, response) {
  console.log('fetchcomplete: ' + queueItem.url);
})
.on("fetchredirect", function(queueItem, parsedURL, response) {
  console.log('fetchredirect: ' + queueItem.url);
})
.on("fetch404", function(queueItem, response) {
  console.log('fetch404: ' + queueItem.url);
})
.on("fetcherror", function(queueItem, response) {
  console.log('fetcherror: ' + queueItem.url);
})
.on("queueadd", function(queueItem, responseBuffer, response) {
  console.log('queueadd: ' + queueItem.url);
})
.on("discoverycomplete", function(queueItem, resources) {
  console.log('discoverycomplete: ' + queueItem.url);
})
.on("complete", function() {
  console.log("complete");
});
```

Is this a known problem, or am i doing something wrong?

fredrikekelund:
If the entry domain redirects to a different domain, the crawler won't add that URL unless you've added the new domain to `crawler.domainWhitelist` or have permitted the crawler to continue on any domain (by default simplecrawler will only crawl the initial domain, with the exception of any www subdomain). Redirects between pages on the same domain should be added correctly. Your code seems to work as expected (but I suspect you were trying to crawl a different domain than microsoft.com).

If you want to allow initial redirects to a different domain, you could add some custom logic that is triggered on fetchredirect events. This example checks if this was the initial request, and in that case adds the hostname of the new URL to the crawler's domain whitelist.

```js
crawler.on("fetchredirect", function (queueItem, parsedUrl, response) {
	var url = require("url");

	if (crawler.queue.length === 1) {
		crawler.domainWhitelist.push(url.parse(parsedUrl).hostname);
	}
})
```

PR #284, opened 2016-05-25
Fix redirect from HTTP to HTTPS

rgladwell:
Fixed bug caused by redirect from HTTP to HTTPS. In this case then protocol is updated, but simplecrawler attempts to crawl via the previous redirected URLs port (previously 80) causing a 404/410 error to be returned by some servers.

This was previously reported as follows:

https://github.com/cgiffard/node-simplecrawler/issues/145
https://github.com/cgiffard/node-simplecrawler/issues/208

Test cases are difficult to compose for this case because it requires a working TLS/SSL server in the test code.

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -1162,7 +1162,6 @@ Crawler.prototype.handleResponse = function(queueItem, response, timeCommenced) 
 
     // We should just go ahead and get the data
     } else if (response.statusCode >= 200 && response.statusCode < 300) {
-
         queueItem.status = "headers";
 
         // Create a buffer with our response length
@@ -1218,6 +1217,11 @@ Crawler.prototype.handleResponse = function(queueItem, response, timeCommenced) 
 
         if (crawler.allowInitialDomainChange && crawler._isFirstRequest) {
             crawler.host = parsedURL.host;
+        }
+
+        // switch port on redirect to TLS/SSL
+        if (parsedURL.protocol === "https") {
+            parsedURL.port = 443;
         }
 
         // Clean URL, add to queue...


fredrikekelund:
Hi, thanks for filing a PR! Seems reasonable to me that the HTTP>HTTPS redirection behavior could have had some issues, great to see you've addressed them! I'll have a closer look at the changes, but could you fix the lint errors first? :)

rgladwell:
Done, rebased.

cgiffard:
Thanks for filing a PR! Would you mind adding a test for the behaviour? I'm happy to help out if you like.

rgladwell:
@cgiffard I think I would need some help on mocking a HTTPS server in the `testcrawl.js`. I'm not sure how you do this?

rgladwell:
@cgiffard any update on this?

cgiffard:
Sorry, I've been absolutely flat out. I hope you don't mind if I come back to this in a couple of days (hopefully it'll be sooner) — I just don't have the bandwidth right now to give your PR the attention it deserves! Really sorry about that. :(

dzafel:
@rgladwell The issue is with `crawler.processURL()` function, I reported it and added a pull request here: https://github.com/cgiffard/node-simplecrawler/pull/288
Your pull request is ok but it only fixes the issue in one place and the `crawler.processURL()` function is still broken and returns wrong data in other places. The solution is very easy, it's just one line of code - the `port` has to be corrected, it sometimes is null instead of integer. Please check my pull request.

rgladwell:
Fixed by #288

PR #288, opened 2016-06-28
https port fix

dzafel:
I didn't check why, but sometimes URIjs returns `protocol() === https` and empty `port()`, so we need to make sure it is 443 for https:// addresses.

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -697,7 +697,7 @@ Crawler.prototype.processURL = function(URL, context) {
     return {
         protocol: newURL.protocol() || "http",
         host:     newURL.hostname(),
-        port:     newURL.port() || 80,
+        port:     newURL.port() || (newURL.protocol() === "https" ? 443 : 80),
         path:     newURL.resource(),
         uriPath:  newURL.path(),
         depth:    context.depth + 1
--- a/test/resourcevalidity.js
+++ b/test/resourcevalidity.js
@@ -284,5 +284,16 @@ describe("Resource validity checker", function() {
 
         });
 
+        it("should make sure the port number is not empty", function () {
+
+            var crawler = makeCrawler("example.com", 3000);
+
+            crawler.processURL("http://www.example.com").port.should.equal(80);
+            crawler.processURL("http://www.example.com:80").port.should.equal(80);
+            crawler.processURL("https://www.example.com").port.should.equal(443);
+            crawler.processURL("https://www.example.com:443").port.should.equal(443);
+
+        });
+
     });
 });


cgiffard:
Hey! Thanks for filing a PR. Would you mind including a test for the change?

dzafel:
@cgiffard I actually just found this pull request: https://github.com/cgiffard/node-simplecrawler/pull/223
Why did you close it? There is an issue with urijs' `protocol()` and `port()` - you can't be 100% sure it works. I did recreate it few times already, but only on websites that are hosted on Windows machines (IIS). The other pull request is probably better than mine, I'd be happy to see it in the next release, thanks.

dzafel:
@cgiffard I just included the test for the change.

fredrikekelund:
Hey @dzafel, sorry about the slow follow up on this one! The changes look good!

But now that you brought it up for discussion, I'm wondering whether it wouldn't be better to simply leave the port undefined if one wasn't explicitly provided? We really shouldn't need to specify a port explicitly if we're targeting the standard one for that protocol. So maybe we should just strip out `|| 80`. Thoughts?

dzafel:
@fredrikekelund hmm, yes I think it should work but, from what I can see, there is currently too many places in the code that require numeric port, for example: https://github.com/cgiffard/node-simplecrawler/blob/master/lib/queue.js#L43 so your solution would break the code. My solution is the simplest one for now and it is pretty critical, I hope to see it in 0.7.1 soon. Also there is a lot to fix about the 443 port, for example this line is hardcoded for port 80 only so https url will always have `:443` added, but it is actually not needed: https://github.com/cgiffard/node-simplecrawler/blob/master/lib/queue.js#L47
It would be cool to see just one function in the code responsible for building url, so we can easily strip out `:80` from `http://` and `:443` from `https://` addresses, for example like here: https://github.com/cgiffard/node-simplecrawler/blob/master/lib/crawler.js#L386. Now I see few places where urls are build (see the link above for sample, and second one: https://github.com/cgiffard/node-simplecrawler/blob/master/lib/queue.js#L83)

fredrikekelund:
I just merged a big PR (#256) that's been in the works for some time. The changes in there made a pretty big difference to how the queue deals with ports when adding items. I also just pushed ed9c010, which removes the defaulting to port 80 in `crawler.processURL`.

We're actually gearing up for a 1.0 release this weekend, so you'll see the changes in a stable version pretty soon! I'll also make sure to give you the appropriate credit in the release notes, thanks for bringing up this issue :+1:

bymaximus:
Sorry for reopen this, can you confirm this mege changes are in NPM? I installed latest NPM simplecrawler and I dont see those changes related to https port

PR #290, opened 2016-07-08
Let constructor take a single argument: initialURL

fredrikekelund:
All the URL properties we need can (naturally) be infered from
this single argument. And since URL's more often appear as an
entire string than as individual parts, it makes a lot of sense
to supply an entire URL to the constructor.

This also helps the issue where previously the crawler didn't take
an argument for initialProtocol, which in many cases would be more
relevant than initialPort. The initial port would most often be
the default one for the initial protocol, but the developer had to
set those properties in a reverse order. When an entire URL is
supplied, it becomes very natural to always supply the protocol,
and only optionally supply the port.

Thoughts @cgiffard?


README.md
```diff
@@ -98,18 +98,24 @@ The standard way of creating a crawler is to call the `simplecrawler`
 constructor yourself and initiate the crawl manually.
 
 ```js
-var crawler = new Crawler("www.example.com");
+var crawler = new Crawler("http://www.example.com/");
 ```
 
-Non-standard port? HTTPS? Want to start crawling at a specific path? No problem:
+The protocol and host that's used throughout the entire crawl are infered from
```
cgiffard: Note to self: fix apostrophe once merged. :)

lib/crawler.js
```diff
@@ -25,44 +25,33 @@ var QUEUE_ITEM_INITIAL_DEPTH = 1;
...
 
 */
-var Crawler = function(host, initialPath, initialPort, interval) {
-    var crawler = this;
-
-    // Data integrity checks
-    if (initialPort && isNaN(initialPort)) {
-        throw new Error("Port must be a number!");
-    }
+var Crawler = function(initialURL) {
```
cgiffard: Let's throw an error here if we get arguments.length > 1.

lib/quickcrawl.js
```diff
@@ -63,7 +63,7 @@ module.exports = function crawl(url, successCallback, failCallback) {
         throw new Error("Can't crawl with unspecified path.");
     }
 
-    var tmpCrawler = new Crawler(url.hostname(), url.path(), url.port() || (url.protocol() === "https" ? 443 : 80));
+    var tmpCrawler = new Crawler(url.href());
```
cgiffard: I'm wondering whether we should just pass uri through, and do the validation in `crawler.js`.

cgiffard:
Looks great! I'm going to merge (unless you beat me to it) and just fix these small issues I brought up on master. The only reason there URI components are broken out, is well, no reason. It was just something I decided to do back at the beginning of time, and never thought about it since! So this change is a no-brainer. :)

cgiffard:
👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

fredrikekelund:
Good to hear you found it as logical as I did :) Regarding quickcrawl, I'm thinking it might be a good idea to remove that feature if we merge this. It'd be so simple that it might be more confusing to keep it around as a separate mode. Especially as initializing the crawler would be an even simpler operation, with less properties having to be set.

cgiffard:
I think if we remove quickcrawl, we should have an alternative way of starting a crawl and immediately attaching event listeners to it.

Like:

    crawler.crawl("http://foo.com/").on("fetchcomplete", ...).on(...)...

So for the moment I'm in favour of keeping it.

cgiffard:
We don't make much of a song and dance about simplecrawler's chainable interface in the readme, but I use it like that a lot, haha.

fredrikekelund:
Fair enough, let's retain a chainable way of initializing the crawler. How about allowing regular initialization without the `new` operator? We add something like this to crawler.js:

```js
if (!(this instanceof Crawler)) {
    return new Crawler(initialURL);
}
```

And then you could initialize it like:

```js
var crawler = Crawler("http://example.com").on("fetchcomplete", function () {}) // and so forth
```

cgiffard:
Yeah, I like that. The only other `quickcrawl` feature was the ability to attach `fetchcomplete` and `fetcherror` callbacks just by passing them as arguments, and I'm pretty sure nobody used that. It wasn't a clear way to attach them, anyway. :)

fredrikekelund:
Cool! Let me add that in here while I'm at it and then we can go ahead and merge

cgiffard:
Might be worth adding something like:

```js

Crawler.crawl = function() {
    throw new Error(
        "`Crawler.crawl` is deprecated as of simplecrawler/1.0. " +
        "You can now pass a URL directly to Crawler. See the documentation for more information."
    );
};
```

fredrikekelund:
Good idea!

PR #292, opened 2016-07-09
Crawler: Enabled stop to forcibly terminate in-flight requests

### LINKED ISSUE #206: Calling crawler.stop() does not stop the crawling process
I would like to be able to stop the crawler manually. I tried by calling crawler.stop() but nothing happens. Am I doing it wrong?

cgiffard:
##### PR Checklist

---

Fixes #206.

* Added an additional boolean parameter on the call to stop(), which, if called, will forcibly terminate all in-flight requests.
* Requests are stored in the crawler on crawler._openRequests.
* When a request is complete, it is removed from the request array. This logic is repetitious and could stand to do with some consolidation.
* Calling crawler.stop(true) will loop over this request array and calls request.abort() on each.
* The 'fetchstart' event was moved so it would fire after the request initialisation process. Whether this is a good idea or not is up for debate.
* Tests were added.

### Notes

This code could really break crawler operation, and to be honest, I'm not feeling that confident about it. Going over it with a fine-toothed come would be very wise.

### DIFF
--- a/test/reliability.js
+++ b/test/reliability.js
@@ -41,7 +41,7 @@ describe("Crawler reliability", function() {
 
         localCrawler.on("fetchtimeout", function() {
             timesCalled++;
-            localCrawler._openRequests.should.equal(0);
+            localCrawler._openRequests.should.eql([]);
 
             if (timesCalled === 2) {
                 done();
@@ -59,7 +59,7 @@ describe("Crawler reliability", function() {
         localCrawler.queueURL("http://127.0.0.1:3000/img/2");
 
         localCrawler.on("complete", function() {
-            localCrawler._openRequests.should.equal(0);
+            localCrawler._openRequests.should.eql([]);
             done();
         });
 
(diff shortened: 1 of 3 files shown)


cgiffard:
Looks like I need to double check the tests as the base branch has changed! :)

test/reliability.js
```diff
@@ -158,4 +158,73 @@ describe("Crawler reliability", function() {
 
         localCrawler.start();
     });
+
+    describe("when stopping the crawler", function() {
+
+        it("should not terminate open connections unless asked", function(done) {
+            var localCrawler = Crawler.crawl("http://127.0.0.1:3000/");
```
fredrikekelund: Seems like this is what's causing the tests to fail

cgiffard: Don't worry, I saw that. Just didn't know you'd already merged the other change! Haha. :)

fredrikekelund:
Looks great! Let's merge it once you've fixed the tests

test/reliability.js
fredrikekelund: :+1:

cgiffard:
Seeing as it's my PR, you should do the honours — assuming you see no weirdness with it. :)

fredrikekelund:
Just one thing that struck me. The `abort` method that's called on the request, it isn't the same method that gave us grief a while back with the memory leak is it (see #201)? What we have here are requests, and back then we were dealing with response sockets, so we can rely on that the `abort` method actually does abort the request, right?

We should probably just add some documentation for the argument to `crawler.stop` as well!

cgiffard:
Regarding the documentation, I was kinda blocked because I realised we didn't have any documentation of the public methods at all (!!!) — and I wanted to just build the docs from the code comments. I could add something in the interim but I felt like we should probably just do it properly.

cgiffard:
I am *vaguely* worried that this code might break something because it touches some stuff which is really fundamental to the crawler — and particularly finds itself snared in all that code that caused us such grief before. But I don't think the abort method was the one causing memory leaks, unless I'm somehow mistaken.

cgiffard:
I think it might actually be worth leaving this one for 1.01, and making sure we've got a really solid 1.0 release before breaking things! Then if we need to revert it, we've got something pretty atomic and isolated to pull out.

fredrikekelund:
Ah wow, you're right, we don't even mention `crawler.stop` in the README :disappointed: 

Hmm, you sure? It's true that it doesn't really break the public API, so we could include it in a smaller release, but if this affects developers' implementations, they might be happier to just have to look over their code once - ie. when upgrading to 1.0.0.

And like you said, since we haven't got a formal documentation of the API, what's public and what's private is maybe a bit loosey goosey. But on the other hand, a non-enumerable property that starts with an underscore - it should probably be pretty clear that it's intended to be private.

cgiffard:
Well, if you're happy, I'm happy! Merge away. :)

cgiffard:
Once we've got to the JSDoc-ification, I'll put some doc-building thing into the Travis workflow, so we can get our docs automatically generated for the readme.

upugo-dev:
Hey, @cgiffard @fredrikekelund I came across this PR because I need this feature in my implementation of simplecrawler, I'm using crawler.stop(true) and it seems to be working although I have no idea if this is legit as this seems to remain undocumented. 

As an aside, I'm also using a crawler.on('crawlstart') event listener which seems to work, but is undocumented- I found this by searching the code on the repo for "start" because I guessed there would be an event emitter on the crawl start event...

PR #293, opened 2016-07-10
Added concept of download conditions

fredrikekelund:
### PR Checklist

### What this PR changes

It introduces the concept of download conditions.

This new API works in much the same way as fetch conditions, but every callback that's registered is evaluated once the headers for a resource have been fetched. This means the developer has access to the Content-Type and Content-Length etc.

This also means that we're removing the dreaded fetchWhitelistedMimeTypesBelowMaxDepth option AND in turn changing how the maxDepth option behaves. These changes should hopefully make sense to developers that now have very flexible ways of determining which resources to fetch and how.

Since I've added unit tests for download conditions, I've also added unit tests for fetch conditions (that previously didn't have any unit tests)

### Rationale

See #249


test/reliability.js
```diff
@@ -4,17 +4,22 @@ var path = require("path"),
     os = require("os"),
```
fredrikekelund: The reason I changed this file was to compensate for the fact that the new tests I added increased the test suite's total run time

fredrikekelund:
So something I've been thinking about regarding this: would it make more sense to call these new download conditions "fetch conditions" instead, and the old fetch conditions something like "add conditions", or "queue conditions"?

There's kind of a conflict in meaning when adding these and already having a "fetch conditions" concept, when those conditions don't really control fetching, but rather queueing.

What do you think, @cgiffard?

cgiffard:
I'm wondering whether they should actually *be* the same thing, but you can specify where in the request lifecycle the condition takes place, somehow.

🤔

fredrikekelund:
That might make sense actually. But there are also the slight nuances. They're both filtering functions, but aside from that they're applied at different points in time, they also receive different arguments. Are there any examples where a project has used that kind of setup - one filtering concept that can be applied at different hooks? Just to see if there's some good inspiration, I mean

fredrikekelund:
My best idea so far would be to register the filtering functions as event listeners, something like:

```js
crawler.on("fetch", function(queueItem) {
    return false;
});
```

But that feels less intuitive. Although on the other hand, thinking of jQuery and similar libraries, returning false from an event listener means to prevent the default action for that event. But maybe the problem here is that it would only be possible to return false from some event listeners, and it might not be intuitive to tell which ones?

cgiffard:
I kinda like that idea, but I agree it's unintuitive — it's not the default behaviour of EventEmitter and wouldn't be so discoverable. Furthermore it'd mean any checks would continue the existing requirement for synchronicity, and I'm wondering whether we need to move to asynchronous conditions.

But it looks nice.

I guess my best idea so far is something like this:

```js
crawler.addCondition("queue", function(queueItem, queueItemReferrer, callback) {
    checkInMyMagicalDatabase(queueItem, function(err, areWeGoForLaunch) {
        callback(err, areWeGoForLaunch); // Broken out for clarity
    });
});

crawler.addCondition("download", function(queueItem, queueItemReferrer, response, callback) {
    if (somethingSomethingSomething(response)) {
        return callback(null, true);
    }
    
    doSomethingElseAsync(queueItem, function(err, areWeGoForLaunch) {
        callback(err, areWeGoForLaunch); // Broken out for clarity
    });
});
```

That's roughly equivalent to what we have now, but async. But is it actually understandable? Or just familiar?

The other way might be to make a QueueItem into a class, (ugh-*ish*) and attach a method to it like `abortLoading()` or `bailOut()` — that means, wherever it's at in the lifecycle, any event could instruct the crawler to just stop whatever it's doing with that queueItem (or flag it in such a way that it would be ignored by the next stage.) Something like:

```js
crawler.on("fetchheaders", function(queueItem) {
    // Don't load the board reports
    if (queueItem.stateData.contentType.match(/application\/pdf/i)) {
        queueItem.abortLoading();
    }
});

crawler.on("queueadd", function(queueItem) {
    // We keep getting these junky URLs from our content management system
    if (queueItem.url.match(/\&amp\;/i)) {
        queueItem.abortLoading();
    }
});
```

Anyway, I'm not settled on anything yet. Happy to discuss or tease apart other proposals!

The main thing on my mind is that we want a relatively general system, which will enable people to easily implement their own logic without us having to have a whole bunch of really complex branching inside the crawler itself — and that's what download conditions are supposed to do — but what event might be next? At the moment, for instance, we don't fetch 404s, because mostly people don't want the data from 404 pages. But maybe somebody does, and we'll need a way to retrieve them. If we implement our logic as some kind of abstract condition — that could be removed with simple and general configuration, that figurative person who wants to retrieve 404 data will be able to do so without having to file an issue or patch the crawler.

Anyway, something for us to chew on.

cgiffard:
A couple of additional thoughts:

* Async conditions in defined places are easy to handle. But they should time-out — otherwise other people's code could hang the crawler really easily, and we'll get the blame!
* A synchronous method that could be called at any time to abort a queue item is really intuitive, but could suffer from race conditions. Assume the user wants to abort loading an item, but their check is async — we can't just wait around for it without clear instruction. They might come back asking the item to be aborted, but it's already too late (maybe this isn't a problem and they can just filter it out later.)

cgiffard:
Perhaps this is a problem-for-later(tm) and we can merge download conditions now, and have it on the roadmap to consolidate at some later point.

fredrikekelund:
Interesting ideas! I kind of like the idea of expanding queueItems into a class with an abort method. This doesn't really cover the current fetch condition scenario however - except if the developer hooks into the discoverycomplete event, but at that stage the queue iten isn't really being aborted, it's just not being put in the queue.

And while there might be value in async conditions, this is something that could really slow down a crawl if applied to every incoming resource. I also like the simplicity of the current synchronous fetch condition API, it's rather intuitive. 

But yeah, I agree that we should maybe postpone a bigger decision on this. Let's draft some future plans elsewhere and stick to the original plan here!

There's still the naming question though - do we rename the old fetch conditions now that we have these download conditions? As far as making goes, I think "fetch conditions" make more sense for these hooks. The old hooks could be called "queue conditions" maybe. But this also makes upgrading to 1.0 harder, and perhaps that isn't worthwhile just to have a clearer terminology for newcomers. I'm not sure myself..

fredrikekelund:
Just to pool some thoughts on this, @lgraubner @xhmikosr @tmpfs @thecameronjones @hbakhtiyor, do you have an opinion? 

Should we give the name "fetch conditions" to this new concept, and rename the old fetch conditions "queue conditions" or something similar?

Or should we keep the name "fetch conditions" for the current fetch conditions and call this new concept "download conditions"?

doug-nara:
Nice. This looks similar to the change I had in another PR!

lgraubner:
Good work. The names "fetch conditions" and "download conditions" seem fine for me as I can easily imagine what each one is doing.

thecameronjones:
Nice work! The terms fetch and download conditions work great for me. Honestly either way would be understandable.

fredrikekelund:
Alright then, let's keep the name "fetch conditions" for the old ones, and "download conditions" for the new ones! @cgiffard do you want to do the merging?

cgiffard:
I'll have another look at this one today. Just wanna thoroughly explore the possibility of race conditions before going ahead with this — at least from my first look through it seemed good though!

micschro:
I just tried the downloadConditions feature and noticed that the "fetch404" "fetch410" and "fetchError" events won't be emitted if the download is denied, even though the response status is already known when the downloadConditions are evaluated. Is this the intended behavior? 
My use case for download conditions is to make sure that the requests are successful without having to download the bodies of all resources. Right now, I would have to manually inspect the response in a downloadprevented callback to find out whether a request was successful. It seems unexpected to me that the fetch error events aren't emitted if the download is prevented. To my understanding, knowing the response code is a major advantage of a download condition over a fetch condition.

cgiffard:
I think it's worth ensuring that those events are properly emitted, but in the interests of expediency I'd probably like to revisit that after merging this rather than before.

fredrikekelund:
Great catch, @micschro! Would be worth to fix before 1.0.0, but I probably won't be able to fix it for a few days (am on vacation). If you can fix it on master, @cgiffard, then we can merge this first

fredrikekelund:
I pushed a fix for the `fetch404` etc. events issue. Now download conditions are only evaluated if the resource will actually be downloaded otherwise. Once again, thanks for looking into the changes, @micschro!

lib/crawler.js
```diff
@@ -1656,6 +1629,65 @@ Crawler.prototype.wait = function() {
...
+        crawler.removeDownloadCondition(3);
+
+    Returns true if the fetch condition was removed, and throws an error if it
+    could not be found.
+
+*/
+Crawler.prototype.removeDownloadCondition = function(index) {
+    var crawler = this;
+
+    if (index instanceof Function) {
```
cgiffard: Is there a reason we're unable to use indexOf here?

cgiffard:
@fredrikekelund Thanks for merging! Sorry about not getting onto this sooner. I noticed a function where it appears we might be duplicating the functionality of `Array.prototype.indexOf`, but for reasons which were unclear to me — unless there's a specific reason why this was added, I might fix it on master?

cgiffard:
@fredrikekelund Also wanted to thank you for the enormous amount of work you put in here — looks excellent. :)

lib/crawler.js
fredrikekelund: Hmm, you're right, indexOf should work fine. Don't know what I was thinking here :/

cgiffard: We can just change it on master — I didn't see anything else which would be a big issue! :)

PR #299, opened 2016-07-12
Fix race-condition causing incomplete crawls

### LINKED ISSUE #298: Incomplete crawls dues to race-condition on crawl()
If the combination of number of items and fetch time per item is high enough (when compared to the crawler's crawl interval), the crawler can stop crawling before processing the discovery phase.

A simple example of this problem would be:
- Set a low enough crawl interval (say, 1ms)
- Give the initial url to the crawler
- Watch the crawler stop crawling (normally) even though there are still items to be discovered

What is happening behind the scenes:

The crawler (i.e. lib/crawler.js) maintains an array and a counter (`_openRequests` and `_openListeners`, respectively) to know when it is safe to stop crawling (note: `_openRequests` was previously a counter as well).

Since the `fetchcomplete` event is triggered (and the array/counter updated) before reaching the discovery phase, the crawler can reach the next `crawl` step (i.e. the next `setInterval` tick) before having new items added to the queue (since the discovery phase is yet to run, even though the flags have already been updated). When this happens, the crawler (unexpectedly) stops normally.

HugoMFernandes:
### PR Checklist

- [ ] I have included tests for my code

### Rationale
Fixes #298

When the combination of number of items and fetch time was high enough
(when compared to the crawler's crawl interval), the crawler could stop
crawling before processing the discovery phase (in which the items would
be discovered from the latest fetched page).

This commit adds a new counter `_openDiscoveries` that works similarly
to the already existing logic for `_openRequests` and `_openListeners`.
With this, if a `crawl()` is reached while items are still being
discovered (given by `_openDiscoveries`, which is updated on fetch if
the crawl depth is suitable), the crawler will not stop.

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -240,7 +240,8 @@ var Crawler = function(initialURL) {
         _isFirstRequest: true,
         _openRequests: [],
         _fetchConditions: [],
-        _openListeners: 0
+        _openListeners: 0,
+        _openDiscoveries: 0
     };
 
     // Run the EventEmitter constructor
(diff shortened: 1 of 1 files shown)


fredrikekelund:
Again, thanks a lot for reporting this! But I committed a fix that gets at the root of the issue a bit more, see 39f806fa63c589cf3aca39e545c4a8ff8fc73cc8

PR #301, opened 2016-07-13
Only update queue items asynchronously

fredrikekelund:
### PR Checklist

### What this PR changes

The final puzzle piece that enables asynchronous queue
implementations! It adds an `update` method to the queue that
takes a URL, an object that describes the updates to be made to
the queue item, and a callback.

I've also made sure that all the places where the crawler mutates
queueItems that are *in the queue* now use this method.

And of course, I've added some tests.

### Rationale

See #257

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -281,6 +281,8 @@ Crawler.prototype.start = function() {
         return crawler;
     }
 
+    crawler.running = true;
+
     // only if we haven't already got stuff in our queue...
     crawler.queue.getLength(function(err, length) {
         if (err) {
@@ -303,22 +305,20 @@ Crawler.prototype.start = function() {
                 if (error) {
                     throw error;
                 }
+
+                process.nextTick(function() {
+                    crawler.crawlIntervalID = setInterval(function() {
+                        crawler.crawl(crawler);
+                    }, crawler.interval);
+
+                    // Now kick off the initial crawl
+                    crawler.crawl();
+                });
+
+                crawler.emit("crawlstart");
             });
         }
-
-        process.nextTick(function() {
-            crawler.crawlIntervalID = setInterval(function() {
-                crawler.crawl(crawler);
-            }, crawler.interval);
-
-            // Now kick off the initial crawl
-            crawler.crawl();
-        });
-
-        crawler.running = true;
-        crawler.emit("crawlstart");
     });
-
 
     return crawler;
 };
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -1062,98 +1062,97 @@ Crawler.prototype.queueURL = function(url, referrer) {
 
 */
 Crawler.prototype.fetchQueueItem = function(queueItem) {
-    var crawler = this;
+    var crawler = this,
+        timeCommenced = Date.now();
 
-    // Variable declarations
-    var clientRequest,
-        timeCommenced;
+    crawler.queue.update(queueItem.url, {
+        status: "spooled"
+    }, function(error, queueItem) {
+        var client = queueItem.protocol === "https" ? https : http;
+        var agent  = queueItem.protocol === "https" ? crawler.httpsAgent : crawler.httpAgent;
 
-    // Mark as spooled
-    queueItem.status = "spooled";
-    var client = queueItem.protocol === "https" ? https : http;
-    var agent  = queueItem.protocol === "https" ? crawler.httpsAgent : crawler.httpAgent;
+        if (agent.maxSockets < crawler.maxConcurrency) {
+            agent.maxSockets = crawler.maxConcurrency;
+        }
 
-    // Up the socket limit if required.
-    if (agent.maxSockets < crawler.maxConcurrency) {
-        agent.maxSockets = crawler.maxConcurrency;
-    }
+        if (client === https && crawler.ignoreInvalidSSL === true) {
+            client.rejectUnauthorized = false;
+            client.strictSSL = false;
+        }
 
-    // Apply the ignoreInvalidSSL setting to https connections
-    if (client === https && crawler.ignoreInvalidSSL === true) {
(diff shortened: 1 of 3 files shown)


lib/crawler.js
```diff
@@ -1062,98 +1062,97 @@ Crawler.prototype.queueURL = function(url, referrer) {
...
+    var crawler = this,
+        timeCommenced = Date.now();
 
-    // Up the socket limit if required.
-    if (agent.maxSockets < crawler.maxConcurrency) {
-        agent.maxSockets = crawler.maxConcurrency;
-    }
+    crawler.queue.update(queueItem.url, {
+        status: "spooled"
+    }, function(error, queueItem) {
```
cgiffard: How do we handle the error here? :-/

test/queue.js
```diff
@@ -188,4 +209,85 @@ describe("Queue methods", function() {
...
+        crawler.maxDepth = 2;
+
+        function findByUrl(array, url) {
+            return find(array, function(queueItem) {
+                return queueItem.url === url;
+            });
+        }
+
+        crawler.queue.add = function (queueItem) {
+            var storedQueueItem = deepAssign({}, queueItem);
```
cgiffard: I like this. Maybe we can add small `setImmediate` calls to verify that no race conditions are occurring (well, increase the likelihood that we'd catch them here.)

cgiffard:
Like it so far! I'll finish reviewing tonight. Thanks! :)

lib/crawler.js
fredrikekelund: Ah, you're right - there are a few errors we need to handle. Just to clarify a few things about these errors:
I thought initially about retry strategies for the update method and if that was something we ought to have, but I realized that that's only needed for third party queue implementations, and they can handle retry strategies internally. 
So when we do get an error here, do we just throw it or do we emit some kind of event? We can't really go ahead with what we were doing in any case, and the risk of frequent errors from the update method should be very low (I don't even think the default method can return one), so maybe just throw it?

test/queue.js
fredrikekelund: Sure, we could wrap it in some setImmediate calls, but is there any particular scenario where we run the risk of a race condition here.?

cgiffard: No specific reason, but it's worth simulating an asynchronous queue in our tests.

lib/crawler.js
cgiffard: I'm torn between throwing it and emitting it. I think generally we don't want anything to crash the crawler, as there's the potential for lost state. So perhaps emitting an error would be best? Maybe `queueerror` or something?

fredrikekelund: True, throwing the error is a bit like copping out, I guess. What about the fetching process - do we emit an event and then simply abort the fetching? And leave the queue item as it is?

test/queue.js
fredrikekelund: Ah, that's true - let's wrap it in an `setImmediate` call!

lib/crawler.js
cgiffard: I guess if we emit an error, and leave the queue item as it is, at some point the crawler will attempt to pick it up again (and if it's in the process of downloading — that is, the status as held by the queue is something other than `spooled` — it's the client's responsibility to keep an eye on the queue event to make sure that it's being handled appropriately. There's only so much we can do when the error handling strategy for the queue could be, well, anything.)

fredrikekelund: Alright, so we emit an `queueerror` event or something similar, abort the fetching and rely on the queue to pick up the queue item again in the future, at which point it's hopefully recovered and is able to complete the operation. Sounds like a good plan of action!

fredrikekelund:
All done with the async tests and error handling (as well as a test for that error handling)! Hopefully we should be good to merge now, I'll let you have a final look @cgiffard

brendonboshell:
I'm looking at the `Crawler#start` function ([crawler.js#L302](https://github.com/fredrikekelund/node-simplecrawler/blob/asyncQueueItemUpdate/lib/crawler.js#L302)), for instance. It appears to call the `add` method, but then calls `crawl` within the next tick. But, at that point in time, the queue could will be empty (because `add` could be running asynchronously), and it appears the crawl would stop.

Indeed, if I take the existing (synchronous logic) in `FetchQueue#add` and wrap it in a setTimeout, the crawl immediately stops before it even starts.

I haven't check if there are similar issues elsewhere in the codebase.

cgiffard:
Good catch. There's definitely a race condition introduced by gzipping in `master` now, so it might be worth trying to kill two birds with one stone, and reviewing this for more. :-/

I think we need to do something equivalent to `crawler.wait` (or just generalise `crawler.wait` to support this case too) every time we fire up an async function, and resolve it when the response comes back (wait conditions time out, so they seem like a good base to start from.) Using that single method should hopefully prevent race conditions (well, race conditions involving async tasks which take less than the timeout threshold.)

cgiffard:
For reference, the other race condition is described by #298.

fredrikekelund:
@brendonboshell thanks for taking the time to review the changes here! I've updated the code to prevent the race condition (and made sure that the #298 problem remains fixed in this branch as well).

fredrikekelund:
I'd like to try and merge this later tonight unless you have any objections, @cgiffard? I'll fix any merge conflicts in #293 and then go ahead with that one as well, I think

cgiffard:
@fredrikekelund Thanks! This looks great. :)

fredrikekelund:
I realized just after merging this that the crawler.update method shouldn't rely on a URL as the first argument. There can be multiple queue items with the same URL in the queue, so we should have an id property on all queue items after all. In the default queue implementation that could just be the index of the queue item in the queue (but we still need to store it on the queue item as well). This should be a fairly simple change all in all, though!

PR #306, opened 2016-07-21
added Sample event objects section to readme

liesislukas:
i missed preview of objects from events and i had to make fake calls just to get overview of what data is available, so i think it's good for any dev to have these previews in documentation to know what you can get

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ pages and written tens of gigabytes to disk without issue.
 - [Events](#events)
     - [A note about HTTP error conditions](#a-note-about-http-error-conditions)
     - [Waiting for asynchronous event listeners](#waiting-for-asynchronous-event-listeners)
+    - [Sample event objects](#sample-event-objects)
 - [Configuration](#configuration)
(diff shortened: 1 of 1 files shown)


fredrikekelund:
Hi, thanks for filling a PR! There is already [some documentation](https://github.com/cgiffard/node-simplecrawler#queue-items) for the structure of queue items. Do you think we need to complement that documentation with a more concrete example? Or maybe we ought to have more links in the README to that section?

liesislukas:
I think you need more examples. Show structure of objects that are returned by events. 


BTW, i was using npm's version and now i know that github's version is newer and have some changes, i don't know if those are related to structure of returned objects (what i've added to documentation with this PR) but just keep it in mind.

cgiffard:
Thanks. I think the path forward is to include an example object in the queue documentation, but we'll probably do so in a slightly different way from what you've proposed here.

Here's what I think the path forward is:

* Clearly specify the object described in the queue documentation is of type `queueItem`
* Include an example object below the `queueItem` definition.

Issue #258, opened 2016-03-29
Update code inline docs to JSDoc style


fredrikekelund:
I don't know if what we're currently using has a formalized standard, but I think it'd be good to keep more documentation in the code, and possibly it'd be good to be able to automatically generate documentation from it as well, in which case something like JSDoc might be a good candidate.

cgiffard:
The standard is currently TomDoc, which I like. I'm happy to use JSDoc, but if we can get a TomDoc linter, sticking with TomDoc would be *my* personal preference, since to my eye it looks nicer when you're reading the code. But happy to hear counterpoints to that. :)

cgiffard:
In my most recent project I used ngdocs, but found that the formal documentation wasn't hugely useful. Perhaps here it'll be better to be able to autogenerate API docs.

fredrikekelund:
Ah, I see - well TomDoc looks pretty nice, I didn't realize the comments followed a formal standard. The best module I found for parsing TomDoc comments seems to be [biscotto](https://www.npmjs.com/package/biscotto), but that's unmaintained and only supports CoffeeScript. There's a ticket where JS support has been asked for: https://github.com/gjtorikian/biscotto/issues/47, but it looks like biscotto won't add support for JS. There's of course the possibility of forking it, but for our purposes, I think it'd be easier to change the comments and use an existing tool to generate the docs.

Hmm, what formal documentation are you referring to in your second comment? Wasn't quite clear to me. I haven't used ngdocs, but what standard we use isn't super important to me as long as there's a good tool to parse it.

cgiffard:
I'm fine with either. Ngdoc or jsdoc — both seem pretty reasonable.

cgiffard:
I already ran into a situation where this would be really helpful — we don't currently have documentation for any of our public methods (!!!).

Linking a generated method documentation from the readme would be a good way to bootstrap it!

fredrikekelund:
Just wanted to say that this is in fact underway, I am working on it! It's somewhat tricky to get everything right though, and actually generating a site from the docs will likely highlight some places where we need to change the format of the documentation. But that's an exercise for after the 1.0.0 release

PR #309, opened 2016-07-22
Convert inline documentation to JSDoc

fredrikekelund:
### What this PR changes
It introduces JSDoc documentation format for documenting methods, events and class members. This is still a **work in progress** (only the main crawler module is done), but given the magnitude of the changes here, I wanted to open a PR so we can iron out any issues as early as possible.

I've tried generating a documentation site from this and it actually turned out quite nicely!

### Rationale
See #258

### DIFF
--- a/lib/cookies.js
+++ b/lib/cookies.js
@@ -1,63 +1,49 @@
-/*
- * Simplecrawler - Cookie Jar Functionality
- * https://github.com/cgiffard/node-simplecrawler
- *
- * Copyright (c) 2011-2015, Christopher Giffard
- *
+/**
+ * @file simplecrawler's cookie jar module
  */
 
 var EventEmitter = require("events").EventEmitter,
     util = require("util");
 
-/*
-    Public: Constructor for the cookie jar.
-
-    Examples
-
-        var cookieJar = new CookieJar();
-
-    Returns the cookie jar object which has now been constructed.
-
-*/
-function CookieJar() {
-    var cookies = [];
-    this.__defineGetter__("cookies", function() {
-        return cookies;
-    });
-
-    // Run the EventEmitter constructor
+/**
+ * Creates a new cookie jar
+ * @class
+ */
+var CookieJar = function() {
     EventEmitter.call(this);
-}
+
+    /**
+     * The actual jar that holds the cookies
+     * @private
+     * @type {Array}
+     */
+    this.cookies = [];
+};
 
 util.inherits(CookieJar, EventEmitter);
 
-/*
-    Public: Adds a new cookie to the jar, either by creating a new Cookie() object
-    from specific details such as name, value, etc., accepting a string from a
-    Set-Cookie header, or by passing in an existing Cookie() object.
+/**
+ * Called when {@link CookieJar#add} returns a result
+ * @callback CookieJar~addCallback
+ * @param {Error|null} error   If the operation was successful, this will be `null`. Otherwise it will be the error that was encountered.
+ * @param {Cookie|null} cookie The cookie that was added to the jar
+ */
 
-    name                -   The name of the cookie to add. Alternately, set-cookie
-                            header as string, or an existing cookie object.
-    value               -   The value of the cookie.
-    expiry              -   Expiry timestamp in milliseconds.
-    path                -   Limit cookie to path (defaults to "/")
-    domain              -   Limit cookie to domain
-    httponly            -   Boolean value specifying httponly
-    cb                  -   Optional callback.
-
-    Emits
-
-        addcookie       -   Emitted with new cookie object as an argument.
-
-    Examples
-
-        cookieJar.add("mycookie","myValue",Date.now(),"/","test.com",false);
-
-    Returns the cookie jar object for chaining.
-
-*/
-CookieJar.prototype.add = function(name, value, expiry, path, domain, httponly, cb) {
-
+/**
+ * Adds a new cookie to the jar, either by creating a new {@link Cookie} object
+ * from specific details such as name, value, etc., accepting a string from a
+ * Set-Cookie header, or by passing in an existing {@link Cookie} object.
+ * @fires CookieJar#addcookie
+ * @param {String} name                       Name of the new cookie
+ * @param {String} value                      Value of the new cookie
+ * @param {String|Number} expiry              Expiry timestamp of the new cookie in milliseconds
+ * @param {String} [path="/"]                 Limits cookie to a path
+ * @param {String} [domain="*"]               Limits cookie to a domain
+ * @param {Boolean} [httponly=false]          Specifies whether to include the HttpOnly flag
+ * @param {CookieJar~addCallback} [callback]
+ * @return {CookieJar}                        Returns the cookie jar instance to enable chained API calls
+ */
+CookieJar.prototype.add = function(name, value, expiry, path, domain, httponly, callback) {
     var existingIndex = -1, newCookie;
 
     if (arguments.length > 1) {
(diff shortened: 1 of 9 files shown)


fredrikekelund:
I pushed some documentation for the queue module. I'll try and push some for the cookie module tomorrow as well, but after that I think we should try and add additional documentation incrementally. What we have at this point will help both ordinary developers who want to use simplecrawler as well as those who would like to write third party queue implementations

fredrikekelund:
Alright, that's the cookie module documented as well! The diff here is pretty big by now, and I've snuck some changes in functionality in there as well. I'll try to leave a line comment on all of those to highlight them if you want to have a look, @cgiffard

lib/cookies.js
```diff
@@ -1,63 +1,49 @@
...
-
-    Examples
-
-        cookieJar.add("mycookie","myValue",Date.now(),"/","test.com",false);
+    /**
+     * The actual jar that holds the cookies
+     * @private
+     * @type {Array}
+     */
+    this.cookies = [];
```
fredrikekelund: Was there a particular reason for the custom getter setup with the cookies array previously? Otherwise, this format worked right off the bat and makes it a bit easier to document and understand

lib/cookies.js
```diff
@@ -70,130 +56,131 @@ CookieJar.prototype.add = function(name, value, expiry, path, domain, httponly,
...
+ */
+CookieJar.prototype.remove = function(name, domain, callback) {
     var cookiesRemoved = [],
         jar = this;
 
     jar.cookies.forEach(function(cookie, index) {
-
         // If the names don't match, we're not removing this cookie
-        if (Boolean(name) && cookie.name !== name) {
+        if (typeof name === "string" && cookie.name !== name) {
```
fredrikekelund: Thought I'd express intent a bit clearer here

lib/crawler.js
```diff
@@ -50,124 +40,276 @@ var Crawler = function(initialURL) {
...
-    // we've fetched from
-    crawler.touchedHosts = [];
-
-    // Supported Protocols
-    crawler.allowedProtocols = [
+    /**
+     * Controls which URL to request first
+     * @type {String}
+     */
+    this.initialURL = initialURL;
```
fredrikekelund: Assigning properties directly to `this` rather than to a reference to `this` (ie. `crawler`) makes it so that JSDoc can understand what's a class member

lib/crawler.js
```diff
@@ -297,39 +465,34 @@ Crawler.prototype.start = function() {
...
-
-    Returns a boolean
+/**
+ * Determines whether robots.txt rules allows the fetching of a particular URL
+ * or not
+ * @param  {String} url The full URL of the resource that is to be fetched (or not)
+ * @return {Boolean}    Returns true if the URL is allowed to be fetched, otherwise false
  */
-Crawler.prototype.urlIsAllowed = function (queueItem) {
+Crawler.prototype.urlIsAllowed = function(url) {
```
fredrikekelund: I think this was my attempt to make this method more flexible at some point. Coming back to it now, though, I just found it to make the method more confusing. And when calling this method, you're always going to have access to the queue item's URL if you have access to the queue item itself

lib/crawler.js
```diff
@@ -1380,21 +1474,27 @@ Crawler.prototype.handleResponse = function(queueItem, response, timeCommenced)
         queueItem.status = "redirected";
 
         // Parse the redirect URL ready for adding to the queue...
-        referrerQueueItem = crawler.processURL(response.headers.location, queueItem);
+        redirectQueueItem = crawler.processURL(response.headers.location, queueItem);
 
-        // Emit redirect event
-        crawler.emit("fetchredirect", queueItem, referrerQueueItem, response);
```
fredrikekelund: This used to be plain wrong. `referrerQueueItem` was always the queue item for the next request, so I fixed the wording here

lib/crawler.js
```diff
@@ -1096,17 +1154,19 @@ Crawler.prototype.fetchQueueItem = function(queueItem) {
...
-        clientRequest.abort();
     });
 
-    clientRequest.on("error", function(errorData) {
-
-        // This event will be thrown if we manually aborted the request,
-        // but we don't want to do anything in that case.
-        if (clientRequest._crawlerHandled) {
+    clientRequest.on("error", function(error) {
+        if (clientRequest.aborted) {
```
fredrikekelund: We can rely on this property set by node itself when a request is aborted rather than a manual one. It expresses intent more clearly as well

lib/crawler.js
```diff
@@ -224,28 +383,44 @@ var Crawler = function(initialURL) {
...
     var hiddenProps = {
-        _robotsTxts: [],
+        _fetchConditions: [],
         _isFirstRequest: true,
+        _openListeners: 0,
         _openRequests: [],
-        _fetchConditions: [],
-        _openListeners: 0
+        _robotsTxts: [],
+        _touchedHosts: []
```
fredrikekelund: I made `touchedHosts` a "private" member, since it really is meant only for internal use by the crawler

lib/queue.js
```diff
@@ -41,21 +30,79 @@ function compare (a, b) {
     return matches;
 }
 
+/**
+ * Creates a new queue
+ * @class
+ */
 var FetchQueue = function() {
-    this.oldestUnfetchedIndex = 0;
```
fredrikekelund: I made `referrerQueueItem`, `completeCache` and `scanIndex` non-enumerable properties and renamed them to start with an underscore. We don't use them when defrosting a queue, so it makes sense to not include them in the frozen JSON file either (and this is a good way of avoiding that)

lib/queue.js
```diff
@@ -41,21 +30,79 @@ function compare (a, b) {
...
+            "requestLatency",
+            "requestTime"
+        ]
+    });
 };
 
-module.exports = FetchQueue;
+FetchQueue.prototype = Array.prototype;
 
-FetchQueue.prototype = [];
```
fredrikekelund: Previously, if for some reasons multiple queue instances would have been created, they would've inherited from the same array. Now they inherit from `Array.prototype` instead

lib/queue.js
```diff
@@ -41,21 +30,79 @@ function compare (a, b) {
...
+ */
+
+/**
+ * Adds an item to the queue
+ * @param {QueueItem} queueItem             Queue item that is to be added to the queue
+ * @param {Boolean} [force=false]           If true, the queue item will be added regardless of whether it already exists in the queue
+ * @param {FetchQueue~addCallback} callback
+ */
 FetchQueue.prototype.add = function(queueItem, force, callback) {
-    callback = callback instanceof Function ? callback : noop;
```
fredrikekelund: In the current version of the queue, callbacks aren't optional, so I've removed all of these lines

package.json
```diff
@@ -28,7 +28,8 @@
   "scripts": {
     "lint": "eslint example/ lib/ test/",
     "mocha": "mocha -R spec -t 5000",
-    "test": "npm run lint && npm run mocha"
+    "test": "npm run lint && npm run mocha",
+    "docs": "jsdoc -c jsdoc.json"
```
fredrikekelund: Give it a try! :)

lib/cookies.js
cgiffard: I wish I could remember! I see no reason not to change it now. :)

cgiffard:
In the interests of getting a release out, I'd be keen to pull this off the 1.0 release and push it to 1.1. :)

cgiffard:
(Doing that now, unless there are strong arguments to the contrary!)

(There's no reason we can't get the 1.1 out quickly afterward.)

cgiffard:
Actually, 1.0 looks pretty imminent. Let's just get this over and done with!

fredrikekelund:
Started working on fixing the merge conflicts here yesterday evening, but they were pretty major, so I didn't have the time to finish. Will push the fixes later today!

fredrikekelund:
Alright, I fixed the conflicts and fixed some stuff I had previously missed. Think we're good to merge this, @cgiffard?

FYI, I used the DocBlockr text editor plugin (I think it's available for several editors - at least Sublime and Atom) to help easily generate JSDoc templates for methods and members. If it turns out that it actually is subjective in how it formats comments in some way (although I don't think that's the case), then we know where that came from

lib/cookies.js
cgiffard: FWIW I kinda read this line as "Ah, we're testing it to see if it's a string, so it's probably a polymorphic variable." That said though, it's pretty clearly documented, so I don't see any problem with continuing as is. :)

cgiffard:
I like it. I think we can merge. There are some style things I'd like to tweak a little — adding a bit more vertical whitespace in, and to try and limit line length a little, but they're not impediments for merging! From my perspective, press the button whenever you're ready to go! I'm excited. :)

Thank you again!

lib/cookies.js
fredrikekelund: Think you're right on this one, reverted it!

Issue #278, opened 2016-04-28
Respect meta robots


lgraubner:
As the crawler has now an option to respect `robots.txt` it should also be possible to check the `<meta name="robots" content="noindex,nofollow">` tag.
I'm not sure if it's possible/wanted to process the `noindex` part, but the `nofollow` could be respected.

cgiffard:
First of all, thanks for filing an issue — and apologies for the delay in my response! I just started a new gig and it's been really full on — haven't had much time for github in the last month.

This is one of those ones that I think is a really good idea, but I wonder if we can fully respect it without including a proper HTML parser. The `<base>` attribute is the same — from a user perspective it's definitely sensible for us to include support.

There's probably a broader conversation on the direction of `simplecrawler` in general. When it started it was more of a toolkit for building crawlers, but it's slowly becoming a fully featured crawler in its own right (not necessarily a bad thing, but we should be clear on a direction, and then work toward that philosophy.)

I think, given the number of issues we get asking for simplecrawler to support various HTML annotations, that it might be worth incorporating cheerio (and maybe even passing the cheerio document to event listeners.) But I'm happy to be swayed either way. Any thoughts, @fredrikekelund?

lgraubner:
Thanks for your reply. I was thinking about the robots meta tag as the `robots.txt` will be incorporated in the next release and this makes it kinda complete. In general I like the concept of a basic API which can be customized as there are various use cases.
I'd love to see a pluggable option. Something like the base API (as it exists) and a solid "npm install and you are ready to go" crawler. For that case this would have to be seperated explicitly as I already said in #277.

cgiffard:
I don't mind incorporating cheerio, as it's a pretty small dependency (as things go.) But generally the philosophy is to keep things lean.

lgraubner:
I'd love to contribute. The following `discoverRessources` function I'm using for a project relying on `simplecrawler`. It works with `base` tags, robots meta could be easily added I guess.

```JavaScript
function discoverRessources(buffer, queueItem) {
  var $ = cheerio.load(buffer.toString('utf8'));

  // parse links
  var links = $('a[href]').map(function () {
    var href = $(this).attr('href');

    // exclude "mailto:" etc
    if (/^[a-z]+:(?!\/\/)/i.test(href)) {
      return null;
    }

    // remove anchors
    href = href.replace(/(#.*)$/, '');

    // check if link is relative
    if (!/^https?:\/\//.test(href)) {
[...]

PR #315, opened 2016-08-05
Crawler#respectRobotsTxt now also looks for nofollow robots meta tags

fredrikekelund:
### What this PR changes
If the default Crawler#discoverResources method finds a <meta name="robots" value="nofollow"> tag, it will return an empty array.

### Rationale
See #278

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,10 @@ change to adapt it to your specific needs.
     Controls whether the crawler should respect rules in robots.txt (if such a
     file is present). The
     [robots-parser](https://www.npmjs.com/package/robots-parser) module is used
-    to do the actual parsing.
+    to do the actual parsing. This property will also make the default
+    `crawler.discoverResources` method respect
+    `<meta name="robots" value="nofollow">` tags - meaning that no resources
+    will be extracted from pages that include such a tag.
 * `crawler.queue` -
     The queue in use by the crawler (Must implement the `FetchQueue` interface)
 * `crawler.allowInitialDomainChange=false` -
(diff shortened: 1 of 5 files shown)


fredrikekelund:
@cgiffard I've updated the 1.0.0 release note draft to include this change. I think we're good to go with the release now :shipit: I'll go ahead and publish tomorrow unless you have any objections?

cgiffard:
@fredrikekelund Sounds good! :)

fredrikekelund:
@cgiffard hehe well if you respond that promptly, then I'll go ahead and release straight away!

fredrikekelund:
Aaand done! :beers: :dancer:

cgiffard:
@fredrikekelund Awwwwww yeah. :) Thanks!

I've got a free weekend coming up (finally!) so I'm hoping to get the CLI polished up, and maybe start on a github pages thing, where we can also stuff our generated docs. :)

fredrikekelund:
Ah, nice to hear! :) Would be awesome if we could intertwine the API docs with some more descriptive text content (like we have in the README right now).

I've always found Sequelize's docs to be great, and they do just that - auto-generate docs that contain more than just the global classes. I think they use the mkdocs and dox modules. Seems they have a bit of a custom setup to glue everything together, but it's good inspiration in any case

cgiffard:
That would be good to look at. I'm thinking we can probably inline usage examples too.

fredrikekelund:
I think we should definitely aim for that! Descriptive text and usage examples in more or less the same place as the raw API documentation. As for Sequelize, https://github.com/sequelize/sequelize is the repo, and http://docs.sequelizejs.com/en/v3/ is the docs site

PR #325, opened 2016-09-10
Crawler.discoverRegex: Fixed srcset detection

binarystash:
## PR fixes srcset detection

* Srcset attributes of multiple images are detected.
* Extra attributes after srcset are filtered out.
* Xmldom instead of regex is used to extract the attribute.


binarystash:
I overlooked xmldom warnings during testing. Please ignore this PR.

PR #326, opened 2016-09-10
Crawler.discoverRegex: Fixed srcset detection

binarystash:
## Crawler.discoverRegex: Srcset detection was fixed.

* Srcset attributes of multiple images are detected.
* Extra attributes after srcset are filtered out.
* Xmldom instead of regex is used to extract the attribute.

## Notes

* Xmldom warnings are suppressed in this PR.

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -15,7 +15,8 @@ var http            = require("http"),
     zlib            = require("zlib"),
     util            = require("util"),
     iconv           = require("iconv-lite"),
-    robotsTxtParser = require("robots-parser");
+    robotsTxtParser = require("robots-parser"),
+    DOMParser       = require("xmldom").DOMParser;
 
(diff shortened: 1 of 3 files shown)


fredrikekelund:
Hey, thanks for filing a PR! :) There's been previous discussion around whether or not to integrate HTML parsing modules like cheerio or xmldom into simplecrawler. Our philosophy so far has been to maintain a lean base and not bring in any extra dependencies, but rather we let the end developer easily swap out the `Crawler#discoverResources` method.

Because of how different the demands on different crawlers are, swapping out that method is standard practice. When I've worked with simplecrawler for bigger systems, I've done so almost every time (and I've brought in an HTML parser in the process). So please don't hesitate to apply these techniques in your specific project - but we're not looking to bring that functionality into the core project ATM.

I'll close this PR, **however**, if you've found the current srcset detection to be buggy, you're very welcome to submit a PR to update the RegExps that are used by default!

PR #351, opened 2017-02-16
Queue.defrost: use Math.min to replace Math.max

### LINKED ISSUE #339: After "freeze" / "defrost" logic crawler suddenly stops and "complete" event is never triggered
Hello, thanks for the great crawler and sorry my poor English.

## What happened?
After "freeze" / "defrost" logic crawler suddenly stops and "complete" event is never triggered

## What should have happened?
Crawling process should finish with the "complete" event after "freeze" / "defrost" logic

Use "freeze" / "defrost" logic

## My explanation
FetchQueue.prototype.defrost function use Math.max(queue._oldestUnfetchedIndex, i) and set queue._oldestUnfetchedIndex. 
For example, if  "freeze" queue consist of 4 items: "downloaded, queued, queued, queued" it will return queue._oldestUnfetchedIndex = 3. Then two previous "queued" items will be lost forever and "complete" event will never be triggered with (completeCount === length) statement. 

## My suggestion
Use Math.min(self.oldestUnfetchedIndex, index) logic  as in previous versions with "defrost" function

## Tmp solution
Use queue._oldestUnfetchedIndex = 0 in "defrost" function callback

civalin:
Fixes #339

This patch try to fix `queue._oldestUnfetchedIndex` calculate error
when `queue.deforst(...)` call. This bug causes queue._oldestUnfetchedIndex
always point to the newest element but not oldest one and skip all
previous queued elements.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -464,7 +464,7 @@ FetchQueue.prototype.defrost = function(filename, callback) {
             return callback(error);
         }
 
-        queue._oldestUnfetchedIndex = 0;
+        queue._oldestUnfetchedIndex = defrostedQueue.length - 1;
         queue._scanIndex = {};
 
         for (var i = 0; i < defrostedQueue.length; i++) {
@@ -472,7 +472,7 @@ FetchQueue.prototype.defrost = function(filename, callback) {
             queue.push(queueItem);
 
             if (queueItem.status !== "downloaded") {
-                queue._oldestUnfetchedIndex = Math.max(queue._oldestUnfetchedIndex, i);
+                queue._oldestUnfetchedIndex = Math.min(queue._oldestUnfetchedIndex, i);
             }
 
             queue._scanIndex[queueItem.url] = true;


civalin:
Btw, I think (maybe another related bug) the [L474](https://github.com/cgiffard/node-simplecrawler/pull/351/commits/2287b9c850f754d609613fdf758a06e3fa4c35fb#diff-b68fbb0f945cdf9cddec65ddf0fbe456R474) may also be changed from:

    queueItem.status !== "downloaded"

to:

    queueItem.status !== "downloaded" || queueItem.status !== "redirected"

But not sure this logic is correct or not. Please help to check it out. Thanks!

fredrikekelund:
Hi @civalin, thanks for looking into this and submitting a PR! 👍 It seems you're right that the status check should be updated as well - we should simply check if `queueItem.status === "queued"`. I'll update that and cut a new release to npm

civalin:
Great! Thank you. :+1:

PR #357, opened 2017-03-07
Fix deleting requestOptions.port when blank/no port set

stilliard:
## What this PR changes
Deletes requestOptions.port when default blank / no port returned, same as it does when port 80 or 443 specified. 

## Rationale
Currently the getRequestOptions() method in crawler.js deletes the port setting when it's set to a default 80 or 443, but I noticed when connecting in a different https agent that the default port is often blank which should also be deleted here I believe. 

This then fixes also an issue an issue I experienced when using `proxy-agent` as the httpAgent & httpsAgent where for HTTPS requests the port would end up as 80 when passed over as blank. 

Hope this makes sense, thanks.

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -536,7 +536,7 @@ Crawler.prototype.getRequestOptions = function(queueItem) {
     }
 
     // If port is one of the HTTP/HTTPS defaults, delete the option to avoid conflicts
-    if (requestPort === 80 || requestPort === 443 || requestPort === "") {
+    if (requestPort === 80 || requestPort === 443 || !requestPort) {
         delete requestOptions.port;
     }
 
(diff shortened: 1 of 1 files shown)


fredrikekelund:
Hi, @stilliard! Thanks for filing a PR. Sure, this makes sense to me. Would we cover any additional edge cases by doing a more lax comparison, like `if (requestPort === 80 || requestPort === 443 || !requestPort) {` instead of `if (requestPort === 80 || requestPort === 443 || requestPort === "") {`?

stilliard:
Hi @fredrikekelund! Sure, I think it shouldn't be a problem as servers can't bind to port 0 anyway. 
& It'll be a little safer in case in the future the port comes back as null or false instead of a blank string. 
I can make the change in the pull request now if you'd like?

fredrikekelund:
That would be great, I can go ahead and merge this after you've made the change!

stilliard:
Sure no problem, that change has now been made.

fredrikekelund:
Thanks @stilliard!

Issue #364, opened 2017-04-06
Bad Performence in Big Sitemaps (item - discovery)


geramy92:
## What happened?
If you crawl bigger pages then they have mostly a sitemap or linked pages with a huge number of links. In my case we are talking about 400.000 Links on one sitemap. I know that this is not a normal number but it shows a big problem of simplecrawler that is also happen on smaller sitemaps which have around thousands of links. The actual discover functionality of simplecrawler is completely syncronous and blocks the complete nodejs process. Additionally it uses a lot of .map and .reduce functions. Which are quite slow.

I tried to debug it and could found that the main problem is the function "cleanExpandResources"  (see https://github.com/cgiffard/node-simplecrawler/blob/master/lib/crawler.js#L817). More specific: The reduce on line 829.

## What should have happened?
This functionality should be replaced by async functions. There are packages or examples (I do not know what is the best here) to do that. So that the main NodeJs Process is not blocked anymore by this sync approach.

## Steps to reproduce the problem
See Text. 
I cannot provide here my sitemap because it is within a private network. But I think it is easily to create such a file by adding a lot of Urls (or just use a google sitemap on the internet that has a lot of urls.)

fredrikekelund:
Thanks for leaving a report, @geramy92! I don't believe that node.js has a concept of multithreading that's flexible enough to support a situation like that. This is a core point about node.js - I/O is asynchronous (multithreaded), but all computation operations run in a single thread.

I'm curious to see what kind of performance hits you're taking with that many links, though. While we could refactor `Crawler#cleanExpandResources` with the aim of improving performance, it is a somewhat complex method and it would be nice to know what kind of gains we stand to make performance wise.

If you have a concrete suggestion for how to accommodate this behavior, I'm all ears! But I think that generally, modules that deal with asynchronicity will be *introducing concepts to help deal with operations that are asynchronous by default*, those modules *won't enable us to simply make code asynchronous*.

geramy92:
I mean packages like https://github.com/jonschlinkert/async-array-reduce or https://github.com/caolan/async  they provide completely custom implementations instead using reduce. 

I didn't try it out right now, but if you want I can try some different options out or try write something custom but they show how you could solve it in this case.

fredrikekelund:
[...]

PR #365, opened 2017-04-10
Change cleanExpandResources to work async

geramy92:
## What this PR changes
- changed cleanExpandResources to work async via multiple ticks (process.nextTick) by change Array.reduce to async.reduce (package was already in project)
- added callbacks and adapted tests for that

## Rationale
related to Issue #364 

This will not improve performence. To do that, we would need a lot of more changes in architecture. But it will stop cleanExpandResources to block the whole process. With that it is possible that a package that uses simplecrawler can do it's stuff and has not to wait over hours. (In Our case we have an interval for logging and such stuff which was blocked before)


geramy92:
seems I have forgotten something....

PR #370, opened 2017-04-24
crawler.urlNormalize option

TimorTartakovsky:
*Thanks for filing a PR with simplecrawler, we appreciate all the help we can get! Please delete this text and fill out the template below. If you haven't read the contributing guidelines, here's a tl;dr to look through before submitting.*

- *Did you make sure the commit messages are clear, and WIP commits have been squashed?*
- *Did you lint and test the code (`npm test`)?*
- *Did you include tests for your code?*
- *If applicable, did you update the documentation with the changes you made?*
- *If you made changes to the docs, did you run a quick spell check?*

## What this PR changes
*Give a brief overview of what you've changed*

## Rationale
*If this has been discussed in an issue already, please reference it here!*

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -44,7 +44,13 @@ var Crawler = function(initialURL) {
     EventEmitter.call(this);
 
     var crawler = this,
+        parsedURL = "";
+
+    if (this.urlNormalize) {
         parsedURL = uri(initialURL).normalize();
+    } else {
+        parsedURL = uri(initialURL);
+    }
 
     /**
      * Controls which URL to request first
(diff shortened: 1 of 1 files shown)


TimorTartakovsky:
Code review is needed.

fredrikekelund:
Hi @TimorTartakovsky, thanks for taking the time to file a PR! I think we need to discuss the rationale behind the changes here though. I see that you noticed #355 as well, which is similar in scope, but about to be closed.

Have you encountered a situation where normalizing the URL actually led the crawler to the wrong URL in the end?

TimorTartakovsky:
The thing is when we have a  URL http://somedomain:8888/path/A&B and after crawler reach this page and adds it to the queue, this page redirects crawler to another one which is the same but encoded like http://somedomain:8888/path/A%26B then crawler normalize second url and it becomes the same as it was in first example. But the web site expecting encoded version then we have unfetched page at the end. I disabled normalize function and everything worked just fine. But I think that it was not a perfect idea to fully unable normalize function. I will work under this question to may be create partial normalization instead of totally unabling urljs.normalization. Thanks

geramy92:
I think the Problem is the following / To explain our problem simply:
Given we have a url:   protocoll://domain.tld:port/somepath?query
And somepath has some encoded stuff like %26 
Then it is always translated to a regular symbol like &

In a some cases webservers differ between a path with such symbols or encoded symbols
So we only need to disable this encoding of somepath in our case...

PR #371, opened 2017-05-03
Change cache backend parameter to a function

abec:
## What this PR changes
The cache backends parameter did not work. It was originally asking for an object, but treating the object like a function. I've changed it to a function that returns an object.


fredrikekelund:
Hey, thanks for this @abec! Sorry about the slow response. I'll get a release out with this fix later today as well

abec:
Thanks man!

PR #376, opened 2017-06-01
Add option to sort Query parameters

HaroldPutman:
## What this PR changes
I added an option to have crawler sort the query parameters before queuing URL's.

## Rationale
The URLs `http://www.example.com/test.jsp?a=1&b=2` and `http://www.example.com/test.jsp?a=2&b=1` in most cases will return the same content and we want to consider them the same URL. We can canonicalize this URL by sorting the parameters. urijs does this in the [equals()](http://medialize.github.io/URI.js/docs.html#equals) method, but normalize() does not sort the parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,9 @@ change to adapt it to your specific needs.
     Or go even further and strip WWW subdomain from requests altogether!
 * `crawler.stripQuerystring=false` -
     Specify to strip querystring parameters from URL's.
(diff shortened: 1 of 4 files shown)


fredrikekelund:
Thanks for the PR, @HaroldPutman! This is actually great, very clever feature to be able to turn on. I see you've added unit tests that look nice and thorough as well :+1: Will push a new release to npm with this change merged in a minute!

Issue #377, opened 2017-07-12
Problems with removeFetchCondition & removeDownloadCondition


venning:
There are a few bugs with `removeFetchCondition` and `removeDownloadCondition`.  I'm submitting a pull request with this issue that resolves those bugs.  Feel free to ignore it or change it or ask me to change it.

## Typo

First, `removeFetchCondition` has a typo on [this line](https://github.com/cgiffard/node-simplecrawler/blob/cc22c5c9b224924d15a0d2ac72a8bcba7c00c8cf/lib/crawler.js#L1916).  The relevant block:

```javascript
if (index instanceof Function) {
    var itemIndex = crawler._downloadConditions.indexOf(index);
    return Boolean(crawler._fetchConditions.splice(itemIndex, 1));
} else if (typeof index === "number") {
    return Boolean(crawler._fetchConditions.splice(index, 1));
}

throw new Error("Unable to find indexed fetch condition");
```

That second line should obviously be `_fetchConditions` not `_downloadConditions`.

## Errors never throw

But that begs the question "Why are the tests passing?"  That's due to a few logic errors in that block (and the corresponding block of `removeDownloadCondition`).

In the test, `itemIndex` gets a value of `-1` from `indexOf` (since the function is not found in `_downloadConditions`) which is then passed to `splice` which interprets it as "one less than the end of the array" which is coincidentally the correct element (since it's a one element array).  It then removes it successfully and returns, providing a **false positive**.

The remove methods' JSDocs both say:

> If the removal was successful, the method will return true. Otherwise, it will throw an error.

Assuming that this is the contract you want to keep, the remove methods are not functioning correctly.  The methods will never throw an error if passed a function.  They will also never throw an error if passed a number.  This is because `splice` will always successfully return an array, even if it's splicing on an empty array or if the index is negative.  To keep the contract, index bounds checks need to be added to both the function parameter block and the index parameter block.

I propose the following for `removeFetchCondition` (and similar for `removeDownloadCondition`):

```javascript
if (index instanceof Function) {
    var itemIndex = crawler._fetchConditions.indexOf(index);
    if (itemIndex !== -1) {
        crawler._fetchConditions.splice(itemIndex, 1);
        return true;
    }
} else if (typeof index === "number") {
    if (index >= 0 && index < crawler._fetchConditions.length) {
        crawler._fetchConditions.splice(index, 1);
        return true;
    }
}

throw new Error("Unable to find indexed fetch condition");
```

[...]

PR #378, opened 2017-07-12
Fixes multiple issues with removeFetchCondition & removeDownloadCondition

venning:
Relevant issue: #377


fredrikekelund:
Hey @venning, thanks so much for your detailed issue report and extensive PR! 🥇 It's great that you took the time to track down and fix the bug, and I think the changed behavior where both methods throw an error when no fetch/download condition is found makes perfect sense.

I made some minor changes by removing the REMOVED_CONDITION constant and by updating the JSDoc comments as you suggested. Other than that, I'll release a new version to npm with these changes ASAP!

PR #380, opened 2017-07-18
Performance tuning for very large sitemaps.

### LINKED ISSUE #364: Bad Performence in Big Sitemaps (item - discovery)

elxa:
Fixes #364

This replaces Array filter, map and reduce calls in the method cleanExpandResources with a simple for loop because they are very slow for large arrays > O(n^2). Speedup observed was 20x.

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -832,52 +832,52 @@ function cleanURL (URL, queueItem) {
  * @param  {QueueItem} queueItem The queue item representing the resource where the URL's were discovered
  * @return {Array}               Returns an array of unique and absolute URL's
  */
-Crawler.prototype.cleanExpandResources = function(urlMatch, queueItem) {
+Crawler.prototype.cleanExpandResources = function (urlMatch, queueItem) {
     var crawler = this;
 
     if (!urlMatch) {
         return [];
     }
+    const URLs = new Set();
+    let URL;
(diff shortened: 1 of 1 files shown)


elxa:
Would you consider raising the minimum required Node.js version to 6.X ? It seems there are 3 issues with my pull request, which would be solved entirely by requiring Node.js >= 6.X:

1. Using const keyword
2. Using let keyword
3. Using Set datastructure

PR #382, opened 2017-07-20
Performance optimization for very large sitemaps.

### LINKED ISSUE #364: Bad Performence in Big Sitemaps (item - discovery)

elxa:
Fixes #364

## What this PR changes
This refactors the cleanExpandResources method in crawler.js to use a single for loop instead of Array.{filter,map,reduce}. The speedup observed was 20x for a sitemap with 20k urls.
To allow usage of const, let and sets Eslint was bumped to ES6 and Node.js < v4 was dropped from the travis/appveyor configuration.

## Rationale
After creating and analyzing several CPU profiles in Webstorm I decided to get rid of Javascripts native array methods filter, map and reduce, because they are very slow for large arrays. Additionally using a set instead of an array for the result of the reduce part makes a lot more sense because we want to keep unique URLs only.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,8 @@
     "root": true,
 
     "env": {
-        "node": true
+        "node": true,
+        "es6": true
     },
 
     "rules": {
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ sudo: false
 language: node_js
 
 node_js:
-    - "0.10"
-    - "0.12"
     - "4"
     - "5"
     - "6"
--- a/README.md
+++ b/README.md
@@ -884,14 +884,10 @@ list below before submitting an issue.
 ## Node Support Policy
 
 Simplecrawler will officially support stable and LTS versions of Node which are
-currently supported by the Node Foundation. We will endeavour to continue to
-support Node 0.10.x — but now that it has fallen out of LTS it is likely we will
-adopt newer JS syntax and APIs which 0.10.x does not support.
+currently supported by the Node Foundation.
 
 Currently supported versions:
 
-- 0.10.x
-- 0.12.x
 - 4.x
 - 5.x
 - 6.x
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,10 +6,6 @@ version: "{build}"
 
 environment:
     matrix:
-        - nodejs_version: "0.10"
-          platform: x86
-        - nodejs_version: "0.12"
-          platform: x86
         - nodejs_version: "4"
           platform: x64
         - nodejs_version: "4"
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -833,6 +833,7 @@ function cleanURL (URL, queueItem) {
  * @return {Array}               Returns an array of unique and absolute URL's
  */
 Crawler.prototype.cleanExpandResources = function (urlMatch, queueItem) {
+    "use strict";
     var crawler = this;
 
     if (!urlMatch) {
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -832,52 +832,48 @@ function cleanURL (URL, queueItem) {
  * @param  {QueueItem} queueItem The queue item representing the resource where the URL's were discovered
  * @return {Array}               Returns an array of unique and absolute URL's
  */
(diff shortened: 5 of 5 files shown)


fredrikekelund:
Hi, @elxa! Thanks for taking the time to write a PR. I'm happy to change the method to use a for loop instead of array methods, but I'm wondering whether the ES6 features are necessary - how much do they contribute to the speed increase?

elxa:
Thanks @fredrikekelund . I have not checked that. To me it just didn't make sense to use an array to store unique URLs, which is why I chose the set datastructure. That latter seems to require ES6 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set
I assume using an array instead of the set would push the complexity to O(n*m) with m <= n, because of the lookup in https://github.com/cgiffard/node-simplecrawler/blob/master/lib/crawler.js#L871 ?
I have a published a small benchmark using 200k items showing huge performance differences on https://jsperf.com/array-vs-set-has-element
The second test case is what the crawler currently does, the third uses the set. This assumes the worst case scenario where all the URLs processed in cleanExpandResources are valid urls.

elxa:
Thanks @fredrikekelund . I have not checked that. To me it just didn't make sense to use an array to store unique URLs, which is why I chose the set datastructure. That latter seems to require ES6 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set
I assume using an array instead of the set would push the complexity to O(n*m) with m <= n, because of the lookup in https://github.com/cgiffard/node-simplecrawler/blob/master/lib/crawler.js#L871 ?
I have a published a small benchmark using 200k items showing huge performance differences on https://jsperf.com/array-vs-set-add-unique-elements-only-2-0
The second test case is what the crawler currently does, the third uses the set. This assumes the worst case scenario where all the URLs processed in cleanExpandResources are valid urls.

elxa:
Dropping Node.js 0.X not only makes sense technically, but also because it is EOL since the end of 2016 https://github.com/nodejs/LTS#lts-schedule1

elxa:
Initially I didn't add a replacement for the 
```javascript
urlMatch.filter(Boolean) 
```
call in [crawler.js](https://github.com/cgiffard/node-simplecrawler/blob/v1.1.4/lib/crawler.js#L843), but tests with a sitemap of 200k URLs (real pages, no dummy pages) showed that the crawler crashes with this error at some point:

```
TypeError: invalid input
    at URI.p.href (***\node_modules\urijs\src\URI.js:1183:13)
    at new URI (***\node_modules\urijs\src\URI.js:70:10)
    at URI (***\node_modules\urijs\src\URI.js:46:16)
    at ***\node_modules\simplecrawler\lib\crawler.js:1734:24
```
I now added
```javascript
if (!URL) {
    continue;
}
```
in the refactored code.

@fredrikekelund Would you merge this including the bump to es6 or would you rather stay on es5 using an npm package to provide support for sets (I guess there are a few options) ?

fredrikekelund:
I've mulled it over, and as long as we can support 4.x, then I think it makes sense to use ES6 - especially for a feature as well suited as this. I will have to look into whether this requires a major version number bump, though. Will have a closer look at the diff later today!

fredrikekelund:
@elxa the diff looks good, I've merged this now! Will look into making a few other changes and then I will release a new major version to avoid any breakage

PR #383, opened 2017-07-25
Prettier formatting

fredrikekelund:
This PR introduces [prettier](https://github.com/prettier/prettier) as a development tool. prettier is a tool for consistently formatting source code, regardless of what it looked like originally. I've used it in other projects and found it enormously helpful. Not only does it save you time in that you don't have to manually format code, but also it eliminates discussions about code formatting.

In this case, I also changed the indent width to 2 spaces instead of 4. This is because prettier places a heavy emphasis on maximum line length. If it can, it will always make lines shorter than the `--print-width` option (80 by default). In certain methods on the `Crawler` class, we have nested callbacks so deeply that lines got very broken up when using 4 spaces for indentation. This change prevents that, but it also aligns us more closely with [Standard JS], and matches the default for prettier.

Please have a look and see what you think, @cgiffard! I think this will make collaboration on the project easier!


fredrikekelund:
Something weird is going on with node 4. Likely it's because of npm 2 rather than node itself. I'm looking into it, but the end diff should remain largely the same, @cgiffard

cgiffard:
Hey @fredrikekelund — will have a squiz at this shortly!

fredrikekelund:
@cgiffard cool! I realized that there are still some JSDoc comments I have to manually adjust (that's the one thing prettier doesn't handle all that well just yet). Other than that, besides looking at the output, I would recommend quickly testing the workflow with prettier. That's where it really shines. There are editor integrations like prettier-atom that makes formatting a file even simpler (remember to tick the checkbox for ESLint integration)

Issue #385, opened 2017-07-26
Crawler crashes on faulty href values


konstantinblaesi:
## What happened?

The crawler crawler crashes because of this uncaught exception:
```
TypeError: invalid input
    at URI.p.href (***\node_modules\urijs\src\URI.js:1183:13)
    at new URI (***\node_modules\urijs\src\URI.js:70:10)
    at URI (***\node_modules\urijs\src\URI.js:46:16)
    at ***\node_modules\simplecrawler\lib\crawler.js:1737:28
    at FetchQueue.oldestUnfetchedItem (***\node_modules\simplecrawler\lib\queue.js:250:13)
    at Crawler.crawl (***\node_modules\simplecrawler\lib\crawler.js:1729:19)
    at ontimeout (timers.js:386:14)
    at tryOnTimeout (timers.js:250:5)
    at Timer.listOnTimeout (timers.js:214:5)
```

The error happens here:

https://github.com/cgiffard/node-simplecrawler/blob/v1.1.4/lib/crawler.js#L1738

## What should have happened?

The crawler should have skipped the faulty URL without crashing.

## Steps to reproduce the problem

Crawl any page that has a faulty href such as
```<a href="http://">http://</a>```

or use this snippet to reproduce the problem:

```
const uri = require('urijs');

let queueItem = {
    url: "http:///"
}

var url = uri(queueItem.url).normalize();
var host = uri({
    protocol: url.protocol(),
    hostname: url.hostname(),
    port: url.port()
}).href();
```

Would you filter bad URLs in [Crawler.cleanExpandResources()](https://github.com/cgiffard/node-simplecrawler/blob/master/lib/crawler.js#L844) ?

PR #387, opened 2017-07-27
Skip faulty URLs such as those missing a hostname or having an invali…

konstantinblaesi:
See #385 

This needs fixing either in URI.js or in simplecrawler. I agree that it makes more sense to do it in URI.js, but I think it will take a lot more effort to refactor their validation code.

This prevents the crawler crashing on pages containing links such as
```
<a href="http://">http://</a><br>
<a href="http://hostname:port/foo/bar">http://hostname:port/foo/bar</a>
```

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -870,8 +870,21 @@ Crawler.prototype.cleanExpandResources = function (urlMatch, queueItem) {
             continue;
         }
 
-        // If we hit an empty item, don't return it
         if (!URL.length) {
+            continue;
+        }
+
+        let parsedUrl = uri(URL);
+
+        // filter broken urls such as http:///
+        if (parsedUrl.protocol().length // has protocol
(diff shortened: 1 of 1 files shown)


konstantinblaesi:
Looks like me trying to keep URLs such as

konstantinblaesi:
Filtering
```
// filter broken urls such as http:///
if (uriInstance.protocol().length // has protocol
    && !uriInstance.hostname().length // has no hostname
    && uriInstance.path().length // has path
) {
    continue;
}
```

konstantinblaesi:
Looks like 30e38b9cad1a86440486d3423446d49b7c7fee70 somehow broken the integration test on node 8.x ?

konstantinblaesi:
Closing as this is currently discussed and pursued in medialize/URI.js#344 and medialize/URI.js#345.

Issue #386, opened 2017-07-26
Headless Chrome/Firefox instead of PhantomJS for example integration


fredrikekelund:
The lead developer for PhantomJS is stepping down and both Chrome and Firefox are introducing a headless mode. We should look into updating the example code for how to integrate a headless browser with simplecrawler!

MrSnoozles:
Agree, would like to see how to combine simplecrawler with a headless browser.

gombosg:
I'm using a headless browser. Basically, after headers are fetched, in the download condition I tell simplecrawler not to download the page and instead tell my headless module to do so.

Not very reliable when downloading 10000s of pages though.

Issue #361, opened 2017-03-17
Look into proxy logic


fredrikekelund:
The current proxy settings don't really prescribe what type of proxy setup is supported, and we determine which http client to use based on the protocol of the queueItem, regardless of how we access the proxy.

I'd like to add some details to the README about what kind of proxy usage we support, clean up the code and potentially add some integration tests.

Related issues: #357, #336

gombosg:
Yes, I had to use a proxy today and it wasn't trivial. This was a useful resource: https://www.vanamco.com/2014/06/24/proxy-requests-in-node-js/

The point is, you either set the http(s).globalAgent (undocumented in simplecrawler) with some tunneling connection e.g. the https://github.com/TooTallNate/node-http-proxy-agent, or you set the useProxy, proxyHostname and proxyPort parameters.

If you set both, you'll get errors back from the proxy. Tunneling connection (with agent) is way better since it supports https, too.

So I agree, this could use some documenting. :)

chrischrischrischrischris1:
@gombosg , do you recall how? May we see an example in code? 

Thank you

gombosg:
Sure:
(TypeScript, replace imports for JS)
```
import http = require('http');
import httpAgent = require('http-proxy-agent');
import https = require('https');
import httpsAgent = require('https-proxy-agent');

function setProxy() {
    let proxy = process.env.http_proxy; // http://123.456.78.9:3128
    if(proxy) {
        let proxyHostMatch = proxy.match(/(?:\d{1,3}.){3}\d{1,3}/);
        let proxyPortMatch = proxy.match(/(?:(?:\d{1,3}.){3}\d{1,3}:)(\d+)/);
        if (proxyHostMatch && proxyPortMatch) {
            http.globalAgent = new httpAgent(proxy);
            https.globalAgent = new httpsAgent(proxy);
            console.log('Using proxy ' + proxy);
        } else {
            console.log("Unknown proxy: " + proxy);
        }
    }
}
```

This doesn't use Simplecrawler's proxy settings, but rather sets up a global proxy agent connection. Set the environment variable HTTP_PROXY before running.

chrischrischrischrischris1:
@gombosg did you change the code in the crawler.js file to then? 

Sorry, i'm a newbie at this proxy/http stuff and an struggling to get your code blended into the crawler. 

Thanks

konstantinblaesi:
@chrischrischrischrischris1 the snippet
```
[...]

Issue #336, opened 2016-11-16
HTTPS & EPROTO error


zero-:
## What happened?
I tried to crawl a site that switches from http to https and I get fetch client errors. This has got something to do with https because it doesn't happen with other than https sites.

## What should have happened?
It should have fetched the page properly.

## Steps to reproduce the problem
I start to crawl `http://add.org` and it gets the homepage correctly because ADD haven't implemented https redirect to the root. All the links on the homepage are HTTPS so the following fetches are under HTTPS. Code is standard boilerplate from the examples and the error I get is `{"host":"add.org","path":"/pab/","port":443,"protocol":"https","uriPath":"/pab/","url":"https://add.org/pab/","depth":2,"referrer":"http://add.org/","fetched":true,"status":"failed","stateData":{"code":600},"id":12},{"code":"EPROTO","errno":"EPROTO","syscall":"write"}`.

I also applied fix from #288 manually because for whatever reason it doesn't seem to exist in NPM or master but all it did was that now port shows 443 correctly. Without the fix port is empty but in either case it fails.

fredrikekelund:
Hey, thanks for filing an issue! That's kind of a special case there with http://add.org. Do I understand it correctly that even if you apply the changes from #288, it still doesn't work correctly?

simplecrawler does consider sites that use HTTP and sites that use HTTPS as different, and it shouldn't really follow links that use HTTPS from a page that was fetched over HTTP. Even though you say that they code you're using is standard boilerplate, would you mind posting a use case so that I can try and reproduce the problem?

zero-:
Yes, I applied the changes and I see from the error that port 443 is applied correctly at least to the error message ;)

Here's the code. My URLs are domains without protocol or subdomain.

    let myCrawler = new Crawler('http://'+url);
    myCrawler.respectRobotsTxt = false
    myCrawler.useProxy = true

    let prox = _.shuffle(proxies)[0]
    myCrawler.proxyHostname = prox[0]
    myCrawler.proxyPort = prox[1]
    myCrawler.proxyUser = prox[2]
    myCrawler.proxyPass = prox[3]
    myCrawler.maxConcurrency = CONCURRENT_FETCH_THREADS
    myCrawler.maxDepth = 2
    myCrawler.scanSubdomains = false
    myCrawler.filterByDomain = true
    myCrawler.start()

fredrikekelund:
[...]

PR #402, opened 2017-09-27
Basic and X509 Auth, Proxy fixes and Puppeteer support for client side generated content

konstantinblaesi:
- *Did you make sure the commit messages are clear, and WIP commits have been squashed?*
- *Did you lint and test the code (`npm test`)?*  Passes for Node.js >= v6.4.0
- *Did you include tests for your code?* No
- *If applicable, did you update the documentation with the changes you made?* No
- *If you made changes to the docs, did you run a quick spell check?* No

## What this PR changes
* optionally use puppeteer for fetch queueItems. This enables crawling of client side generated content (e.g. ajax) and is one solution to #386 . The feature can be enabled by setting `crawler.clientSideGeneratedContent =  true;`
* always sets `Connection: keep-alive`, because we've encountered slow (application) webservers closing connections before the response was fully sent/received. Chrome seems to do the same.
* Implements http basic and x.509 certificate authentication for individual hosts instead
* refactors the proxy crawler implementation, because according it needs to be handled differently depending on the protocol of the queueItem or the proxy. For now I've used https://www.npmjs.com/package/tunnel which supports all combinations and http basic auth. I've only implemented http over httpand https over http in the crawler for now.

## Rationale
* the integration of puppeteer is one way of solving #386 
* the new proxy implementation might also solve #336 . Needs verification.

* as mentioned in #361 we should probably  add some docs about supported proxy configurations

As you can see in https://github.com/konstantinblaesi/node-simplecrawler/blob/c44f0b4ec18e3d7fc653eb114fa150996b9a2dd2/lib/fetch-client-puppeteer.js#L242 there are some ugly hacks to make the current crawler work with the request/response objects returned by puppeteer, because the current crawler has no generic abstractions for these and simply exposes them in different ways.

Puppeteer requires Node.js  >= v6.4.0 or better >= v7.6.0 .
I'd like to do more refactoring/cleanup making use of async/await including adding missing abstractions. Async/await requires Node.js 8.x (becoming LTS in october) unless you do transpiling one way or another.
What do you think about merging some or all of these commits?
What do you think about raising the Node.js version to >= v7.6 (async/await support) ?

### DIFF
--- a/lib/authentication.js
+++ b/lib/authentication.js
@@ -1,0 +1,110 @@
+class AuthenticationManager {
+    constructor() {
+        this._authentications = {};
+    }
+
+    /**
+     *
+     * @param {String} domainName
+     * @return {boolean}
+     */
+    hasAuthFor(domainName) {
+        return Object.keys(this._authentications).length && domainName in this._authentications;
+    }
+
+    /**
+     *
+     * @param {String} domainName
+     * @return {AuthenticationBasic|AuthenticationX509}
+     */
+    getAuthFor(domainName) {
+        return this._authentications[domainName];
+    }
+
+    /**
+     * Set basic auth configuration for the specified domain name
+     * @param domainName
+     * @param username
+     * @param password
+     */
+    setBasicAuth(domainName, username, password) {
+        this._authentications[domainName] = new AuthenticationBasic(domainName, username, password);
+    }
+
+    /**
+     * Set cert auth configuration for the specified domain name
+     * @param domainName
+     * @param certificatePath
+     * @param certificatePassphrase
+     */
+    setX509Auth(domainName, certificatePath, certificatePassphrase) {
+        this._authentications[domainName] = new AuthenticationX509(domainName, certificatePath, certificatePassphrase);
+    }
+}
+
+class Authentication {
+    constructor(type, domainName) {
+        this._type = type;
+        this._domainName = domainName;
+    }
+
+    /**
+     *
+     * @return {String}
+     */
+    type() {
+        return this._type;
+    }
+
+    /**
+     * Which domain name the authentication data belongs to
+     * @return {String}
+     */
+    domainName() {
+        return this._domainName;
+    }
+}
+
+class AuthenticationBasic extends Authentication {
+    constructor(domainName, username, password) {
+        super(Authentication.Types.Basic, domainName);
+        this._username = username;
+        this._password = password;
+    }
+
+    username() {
+        return this._username;
+    }
+
+    password() {
+        return this._password;
+    }
+}
+
+class AuthenticationX509 extends Authentication {
+    constructor(domainName, certFilePath, certPassphrase) {
+        super(Authentication.Types.X509, domainName);
+        this._certFilePath = certFilePath;
+        this._certPassphrase = certPassphrase;
+    }
+
+    certFilePath() {
+        return this._certFilePath;
+    }
+
+    certPassPhrase() {
+        return this._certPassphrase;
+    }
+}
+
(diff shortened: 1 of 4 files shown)


konstantinblaesi:
Tests on Node.js  < v6.4.0 will fail because puppeteer does not support it.
Would you consider merging the puppeteer code maybe with an experimental notice?

fredrikekelund:
Thanks so much for your contribution, @konstantinblaesi! I am quite swamped at the moment, so I won't be able to look more closely at this for the next couple of days. Looking forward to doing so, however!

I also believe this warrants a discussion about the next generation of simplecrawler architecture. I think it's clear that using puppeteer is something we want to do, so discussing a future architecture to support that would be good.

konstantinblaesi:
Enabling the puppeteer feature is a as simple as setting `crawler.clientSideGeneratedContent = true;`, but the feature should be considered experimental because it does not support all of simplecrawlers features/behaviour, yet. What should work is using basic auth, proxies, feching of page's content. All implementation details regarding puppeteer are in `lib/fetch-client-puppeteer.js`.

fredrikekelund:
Just letting you know that I haven't forgotten about this, @konstantinblaesi! Excited to discuss these updates. I hope to pen a longer reply tomorrow so we can get a discussion going!

LeMoussel:
@konstantinblaesi In `crawler.start()`, would not it be better to set `fetchClientPuppeteer` when `crawler.clientSideGeneratedContent = true`  like this 
```
Crawler.prototype.start = function() {
    var crawler = this;

    if (crawler.running) {
        return crawler;
    }

    if (crawler.clientSideGeneratedContent) {
        crawler.fetchClientPuppeteer = new FetchClientPuppeteer(crawler);
    }
```

fredrikekelund:
Finally getting around to this now! 🥇 Lots of good stuff in here, but I think we're going to want to integrate these updates piece by piece.

The updated way of handling **proxies** is great - I haven't actually dealt with the proxy configuration for simplecrawler very much, but this seems like a much more solid way. From what I can tell, it seems to be fully backwards compatible as well, which is good.

The **authentication** logic is great too - I think it makes a lot of sense to let people specify different authentication credentials for different domains, and supporting X509 seems smart.

The change to include the **`Connection: keep-alive`** header also seems good - we should align with what browsers are doing.

I'm also very intrigued by the **puppeteer** implementation, but this brings me to maybe the most important point to discuss. I've been drafting some bigger API updates that should probably become simplecrawler 2.0. Supporting puppeteer would probably be the biggest feature for that version, and this update looks like a really solid base for that. However, seeing as puppeteer requires us to drop support for node 6.x, I would like to see if we can take some more time to get a solid API design with logical separation of concerns before we merge this. I would also prefer if we released puppeteer support through a separate npm module, so that we wouldn't always have a dependency on it.

So, here's my suggestion for how we move ahead: limit the scope of this PR to just the updated proxy logic, the new auth logic and the `Connection: keep-alive` header. Also make sure that all of those changes are supported all the way down to node 4.

We then start work on a development branch in this repository that will become simplecrawler 2. We'll probably have to find a way to discuss API decisions in an issue. I'll see if I can write down my ideas so far.

What do you think about that? I'm really intrigued by all of this, thank you for pushing simplecrawler forward!

gombosg:
Wow, Puppeteer support would be so great! I'm using it myself as an extension, too - after fetching the headers, I'm calling Puppeteer  manually.

robinparisi:
@konstantinblaesi Awesome work, this PR seems very promising! Any thoughts about @fredrikekelund suggestion?

PR #411, opened 2017-10-07
Update README.md

sebworks:
`.queue.defrost` requires a callback, so this PR updates the documentation to reflect that fact.

### DIFF
--- a/README.md
+++ b/README.md
@@ -679,7 +679,9 @@ crawler.queue.freeze("mysavedqueue.json", function () {
     process.exit();
 });
 
-crawler.queue.defrost("mysavedqueue.json");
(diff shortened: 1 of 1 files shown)


fredrikekelund:
Thanks for bringing this up @sebworks! However, I actually submitted a PR just 30 minutes before you yesterday that brings a bigger overhaul to the README and that makes this update obsolete.

The idea is for us to use generated documentation, making it less likely that we will overlook updating the docs, since the documentation code will be written in the source code. You're welcome to look at the updated README (that will hopefully land soon) [here](https://github.com/simplecrawler/simplecrawler/blob/generate-docs/README.md)

PR #432, opened 2018-06-12
Add a replacement in cleanURL

baptistejamin:
This pull request is fixing an unimplemented special char case in URLs.


kbychkov:
Hello @baptistejamin! Thank you for your PR that should fix [this issue](https://github.com/simplecrawler/simplecrawler/issues/439). I'd like to know where did you encountered that character?

baptistejamin:
I don't remember exactly as it happened back in June.

We format thousands of pages everyday and we log every issue we had. This patch fixed several things with URLs.

kbychkov:
Okay, as far as I see the #439 is not the only case. Therefore this fix could make sense. Please update your branch and I'll merge the PR.

PR #454, opened 2019-02-28
Rewrite deprecated use of `Buffer`

kbychkov:
## What this PR changes
The constructor `new Buffer` was [deprecated](https://nodejs.org/docs/latest-v10.x/api/buffer.html) since Node.js 6.0.0. The new APIs `Buffer.from` and `Buffer.alloc` [should be used](https://nodejs.org/docs/latest-v10.x/api/buffer.html#buffer_buffer_from_buffer_alloc_and_buffer_allocunsafe) instead.

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -560,7 +560,7 @@ Crawler.prototype.getRequestOptions = function(queueItem) {
         var auth = crawler.authUser + ":" + crawler.authPass;
 
         // Generate auth header
-        auth = "Basic " + new Buffer(auth).toString("base64");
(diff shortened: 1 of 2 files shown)


kbychkov:
@cgiffard, I've merged the last two PRs with *Rebase and Merge* option in order to keep [the commit history](https://github.com/simplecrawler/simplecrawler/commits/master) clean. But in this case it's hard to find the related PR. Should I merge future PRs in the usual way, what do you think?

cgiffard:
I don’t feel super strongly about it, but my preference would be to keep squashing. Maybe @fredrikekelund would have a better feel. :)

PR #470, opened 2019-06-10
cache: send headers If-None-Match and If-Modified-Since

### LINKED ISSUE #375: Cache is useless
## What happened?

Cache isn't used. I.e. crawler doesn't send If-None-Match or If-Modified-Since headers
https://github.com/cgiffard/node-simplecrawler/blob/master/lib/crawler.js#L524
Thus server never sends 304 and crawler doesn't use cache.

## What should have happened?

Crawler should verify whether there's an item in cache, get it's ETag of Modified-Since, and generate corresponding headers so that server is able to send 304.

deton:
## What this PR changes
This PR adds If-None-Match or If-Modified-Since header to fetch request
when fetching url is in cache and ETag or Last-Modified value is cached.

This PR also updates queue
on received 304 not-modified response from server.
Otherwise, crawler does not emit "complete" event.

This PR resolves #375.

### DIFF
--- a/lib/crawler.js
+++ b/lib/crawler.js
@@ -573,6 +573,19 @@ Crawler.prototype.getRequestOptions = function(queueItem) {
         requestOptions.headers["Proxy-Authorization"] = proxyAuth;
     }
 
+    if (crawler.cache !== null && crawler.cache.getCacheData) {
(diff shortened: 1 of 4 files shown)


kbychkov:
Hi @deton! Thank you for the PR. It's the issue I've found and would like to resolve too. It should definitely be merged.

PR #483, opened 2020-03-30
Fix string substitution in FS cache backend error constructor

Mr0grog:
## What this PR changes

## Rationale

### DIFF
--- a/lib/cache-backend-fs.js
+++ b/lib/cache-backend-fs.js
@@ -196,7 +196,7 @@ FSBackend.prototype.setItem = function(queueObject, data, callback) {
                     // Just overwrite the file...
                     writeFileData(currentPath, data);
                 } else {
-                    throw new Error("Cache storage of resource (%s) blocked by file: %s", queueObject.url, currentPath);
+                    throw new Error(`Cache storage of resource (${queueObject.url}) blocked by file: ${currentPath}`);
                 }
             }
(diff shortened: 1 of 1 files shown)


Mr0grog:
No problem! 😄

Mr0grog:
Just checking in here… it looks like Travis failed setting up one of the test runs — it couldn’t install Node 12: https://travis-ci.org/github/simplecrawler/simplecrawler/jobs/668529736. It probably just needs to be retried (since I’m not a committer here, I don’t have permission to do that).

konstantinblaesi:
@fredrikekelund @kbychkov ping

kbychkov:
Hi! I've restarted the build but the status in the PR wasn't changed. @Mr0grog I suppose if you'll force push the commit it'll trigger the checks again.

Mr0grog:
Pushed!

Mr0grog:
Well, it seems like it’s stuck in the same spot. All green on Travis, just not showing in GitHub. If you need it to show correctly here, I could close and re-create this PR. Otherwise, I’ll just leave this up to you to merge.

kbychkov:
Well, it's something wrong with Travis. I've tried to merge using command line but was also unsuccessful. @Mr0grog let's re-create the PR. I don't see another way at the moment.

Mr0grog:
Retried in #484 and #485; neither seem to be working. :(

PR #484, opened 2020-04-11
Fix format in FS cache backend error constructor (take 2)

Mr0grog:
**This PR was originally filed as #483. I’m recreating it to try and resolve issues with Travis.**

## What this PR changes

## Rationale


Mr0grog:
Closed this PR on account of issues with Travis that don’t seem to be resolvable. Retrying with a new commit hash in #485.

PR #485, opened 2020-04-11
Fix format in FS cache backend error constructor (Take 3)

Mr0grog:
**This PR was originally filed as #483. I’m recreating it to try and resolve issues with Travis.**

## What this PR changes

## Rationale


kbychkov:
I've created a commit with these changes myself and successfully merged it to the master. Not sure what was happening with Travis though :(

Mr0grog:
Thanks for getting it in, at least!

